PR #9, opened 2016-02-25
Feature/initial trial indexer

roll:
### Data flow
- indexing processor creates `trials` index in `warehouse` and generates trial uuid
- `mapper` use `trials` index table as entry point to iterate over all `ready-to-map` trials
- the same mechanism could be used for `persons`, `organisations` etc
### Table: `warehouse.trials`

```
    op.create_table('trials',
        sa.Column('uuid', UUID, primary_key=True),
        sa.Column('updated', sa.DateTime(timezone=True), nullable=False),
        # records column stores path to entity instance
        # for person it should be `labels` column like `register::uuid::label/hash`
        # {euctr::2205f8f98fcf4c58b90c9b9c247e8115,isrctn::de7da83caca24318b961853a052a0f87}
        sa.Column('records', ARRAY(sa.Text), nullable=False, unique=True),
        sa.Column('nct_id', sa.Text, unique=True),
        sa.Column('euctr_id', sa.Text, unique=True),
        sa.Column('isrctn_id', sa.Text, unique=True),
        sa.Column('scientific_title', sa.Text, unique=True),
    )
```

### DIFF
--- a/docker-compose/processor.yml
+++ b/docker-compose/processor.yml
@@ -1,5 +1,17 @@
-process:
+index-euctr:
   image: opentrialsrobot/warehouse
-  command: echo 'no-op'
+  command: python -m processor.trials.euctr
+  environment:
+    OPENTRIALS_WAREHOUSE_URL:
+
+index-isrctn:
+  image: opentrialsrobot/warehouse
+  command: python -m processor.trials.isrctn
+  environment:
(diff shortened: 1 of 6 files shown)


roll:
@pwalsh please take a look

pwalsh:
:+1: ready to go.

PR #14, opened 2016-03-23
Feature/improved mapper

roll:
Many `mapper` improvements, refactoring, concrete registers mappings etc


roll:
@pwalsh please take a look at PR (cc @vitorbaptista)

Overview - https://github.com/opentrials/processors/blob/feature/improved-mapper/docs/mapper/mapper.md

docs/mapper/mapper.md
```diff
@@ -0,0 +1,135 @@
...
+print(existent)
+True
+print(entity)
+{
+    'id': 'abc'
+    'links': [],
+    'facts': ['nct15', ...],
+    'created': ...,
+    'updated': ...
+    'primary_id': 'nct15',
```
pwalsh, line 157: Shouldn't we have `ids` as an array (talked about this before)?

docs/mapper/mapper.md
```diff
@@ -0,0 +1,135 @@
...
+```python
+pipeline = Pipeline(source=warehouse, target=database)
+pipeline.read(table)
+pipelint.write(table, keys, **data)
+...
+```
+
+### Finder
+
+Finder get `filter`, `links` and `facts` about some entity
```
pwalsh: Can you explain `facts` and `links`?

docs/mapper/mapper.md
roll, line 157: `id` is our internal uuid - https://github.com/opentrials/api/blob/master/migrations/20160216120159_create_initial_schema.js - for registers is we have `primary_id` and `secondary_ids` (WHO fileds)

pwalsh, line 157: I know `id` is our internal uuid. I mean `ids` as an array of what we now call `primary_id` and `secondary_id` (I don't know how we define what is "primary" and what is "secondary"). But, I see it is in the WHO Data Set, so fine. I guess it just feels arbitrary, as it implies a hierarchy that we can't guarantee. I'd be inclined to:
- Remove `primary_id`
- Rename `secondary_id` -> `ids`
- Have a convention where `ids[0]` could be considered "primary"

But, we can leave now as we are following the WHO Data Set as we aim to.

docs/mapper/mapper.md
roll: Yes of course I was a little bit shy here.

For now it's a components of deduplication system.

For example trial `facts` is slugified register indentifiers 
and hashed/slugified scientific titles:
`['nct21231', 'isrct23412', 'sa35434525']`

On deduplication staget finder find trials with this facts
(we need to use GIN index here - I read postgres will use index here)
to associate item from some other register with one living in our database.

If there is a match facts will be merged:
`['nct21231', 'isrct23412', 'sa35434525', 'euctr2224']`

The same for the persons for example where facts will be
slugified `phones`, `email` etc (not fully implemented). But for persons
we also check `name` equality. 

Persons with the same name and one of the facts equal is a one person.

Other conceps is `links` - here stored trial uuids for entities like 
persons - if no facts are match but name is the same and persons
used to be part of the same trial - it'is a one person.

It's quite WIP but the best approack I've found for now to have
bullet proof and flexible dedup system for all entities we need.

pwalsh: Please ad this info to the documentation here :).

pwalsh:
Cam merge. Great work!

docs/mapper/mapper.md
roll: It's what I was saying on the spring planning - to have a few fields in `database` entity tables to query for matches instead a full additional tables in `warehouse` like `index_trials` etc. When any problem with this index (unsync) will break our `uuid` emission system.

docs/mapper/mapper.md
roll, line 157: For now it's more questions than answer on this topic - I suppose we have to check with apps (@vitorbaptista) - what they need. I think we will change it sometimes.

Having things like `primary_id` and `primary_register` is helpful on dedup stage (I use it in this PR) . With notnull fileds there is a guarantee it exists.

May be `secondary_ids` to `ids` and store all ids here? If apps don't need to show just secondary ids. Also we use here `dict` instead of `array` to know type of id.

Current situation is not too eloquent it's true.

roll:
Added `facts/links` notes to docs.

Also I've answered about `ids` but I'm not sure where is the answer. In a few words - we'll see it's not a too eloquent situation for now. We should discuss it between the next `database` schema migration with Vitor.

pwalsh:
@roll great, thanks

PR #19, opened 2016-05-01
Feature/gender published data

roll:
https://github.com/opentrials/opentrials/issues/84


roll:
@vitorbaptista please review

processors/gsk/extractors.py
```diff
@@ -19,9 +19,22 @@ def extract_source(record):
...
+
     # TODO: fix
+    # Get registration date
     registration_date = (
         record['first_received'] or datetime.datetime.now().date())
+
+    # Get gender
+    gender = None
+    if record['gender'] in ['Both', 'Male', 'Female']:
+        gender = record['gender'].lower()
```
vitorbaptista, line 31: When a gender can be different than "Both", "Male" or "Female"? It would be good to add a small comment here.

vitorbaptista:
@roll Looks good :+1:

Just a suggestion, you use this pattern in many places:

``` python
foo = False
if bar:
    foo = True
```

That could be simplified to:

``` python
foo = bool(bar)
```

Or, when the condition is an actual test, you could do:

``` python
foo = (bar == "baz")
```

If the condition is small, I think it's cleaner. However, that's just a minor code style change. Feel free to merge the code as is.

roll:
I've removed gender checks - let fail (process will be ok it's only one iteration and it will be in Papertrial) if something will change in warehouse. So I'll be notified about a problem.

About conditions:

```
foo = False
if bar:
    foo = True
```

I prefer this style because it's more like pseudo-language (could be understood by anyone) and more importantly there is no if branch hiding so code coverage tool will flag about untested condition (almost not applicable here of course).

PR #21, opened 2016-05-02
Bug/record source_data

roll:
Fixes https://github.com/opentrials/opentrials/issues/94


vitorbaptista:
Cool! Didn't realize that we could fix that bug in the processors. Have you tested that this fixes that error? Also, why are you dumping as JSON and then loading? Couldn't we simply pass the original `record`, or that's that needed to convert Python objects (like Dates)?

roll:
Yea, problem with some python types (dumps-loads) - https://github.com/opentrials/processors/blob/master/processors/base/helpers.py#L34

It fixed the error. Sorry I didn't realize Postgres doesn't parse string if it's passing to JSONB field or raise an error. So before fix we silently get JSONB with string as a value here (it wasn't a plan=).

PR #22, opened 2016-05-04
Feature/diffs processor

roll:
https://github.com/opentrials/opentrials/issues/50

(recreated PR after wrong merging)


vitorbaptista:
As a clarification, are those diff simply to see the difference between each `trialrecord`, or they're a log of the changes to the `trial`?

If they're about seeing the difference between each `trialrecord`, could you explain your reasoning of solving this on the DB/processor level? It doesn't seem hard to generate the diffs at the API or even frontend level, but I might be missing something.

roll:
It doesn't preserve state/logs etc.

Design is based on - https://github.com/opentrials/opentrials/issues/50

From our user stories (is Ben's meeting on Tuesday or Thursday? you could ask him how it could be looking):

> As a Visiting User, when viewing a single trial, I want to see, where it exists, the multiple records for a trial side by side, so that I can view these for any interesting discrepancies.
> As a Visiting User, when viewing a single trial which has multiple records, I want a clear UI element that tells me if the trial has discrepancies across these records, so I can investigate these further if they exist.
> **As a Visiting User, I want a search option to filter by trials with known discrepancies, so I can investigate these trials as a group**.

As a result we need to pre-process all diffs here. So to filter all trials by knows discrepancy will be as simple as inner join `trials` with `diffs` table (type of discrepancy like `has_published_results` could be selected).

vitorbaptista:
OK, so there're two functionalities: finding trials with discrepancies and diffing trials. Finding discrepancies must live in the DB or ElasticSearch. If it's in the DB, we could create views (which have their performance issues [and some solutions](https://hashrocket.com/blog/posts/materialized-view-strategies-using-postgresql)). Better yet, we could simply cache in ElasticSearch.

The actual diffing should be done in the frontend, IMHO, considering that the records have their columns normalized.

The issue I see with this solution is that we'll end up having to be extra careful to update the diffs whenever the trials are updated (not to mention the extra storage). We would have to do this as well with ElasticSearch, but there we already have to do it anyway. The query for caching in ElasticSearch could be as simple as:

``` sql
SELECT trial_id, COUNT(*) AS count FROM (
  SELECT DISTINCT ON (public_title) trial_id FROM trialrecords
) AS foo
GROUP BY trial_id;

               trial_id               | count 
--------------------------------------+-------
 05cc77ad-5575-4c04-9309-4c64d5144b07 |     1
 008225d1-2d6f-4437-a6e9-5eb3b738bcc6 |     2
```

Then we just need to filter on `count > 1` and we'll have a list of trials with discrepancies.

What do you think?

roll:
Always wanted to ask - how our ES re-index works?
Because we also have this - https://github.com/opentrials/opentrials/issues/41

roll:
About diff I'm still more leaning for approach proposed by Paul - because report generation generally is separated part of a system because it could include some non trivial business logic. For now diff script is simple but then someone tell us to add some filtering, ignore euctr matches when something = .. and only on Monday etc.

About unsync concern - I don't think it will be a problem - we will have many processor scripts here and all of them will be working consistently processing changed data. I suppose we will finish we some linkers/unlinkers, user input checkers, to warehouse movers, exporters etc. But this component (`processors`) responsibility is process data so I think is ok. Better move here as much as possible to simplify components not responsible of data processing.

roll:
Also I think we need more information from domain specialists here (use cases) to understand what way to choose.

vitorbaptista:
About ES reindexing, we're currently doing it as dumb as possible: delete the index and recreate it. This is the easiest way, but have many drawbacks (e.g. the users getting partial data when the index is being rebuilt). We have to improve it moving forward.

For diffing, I'm concerned that this approach would be difficult to change because the processors are on the lower levels of our architecture. Any changes here will propagate to the API and the Explorer.

My current idea is to change this PR to simply cache which trials have discrepancies in their trialrecords (probably with a query similar to https://github.com/opentrials/processors/pull/22#issuecomment-216884689, maybe simply adding a new bool column to the `trials` table), so we can build the filter. For diffing, we could create an API endpoint like `http://api.opentrials.net/trials/{trial_id}/records` which returns all data for all trialrecords for that trial, then we can diff in the Explorer side.

This keeps the business logic in the Explorer, which is the easiest place to change IMHO. If we need to add custom reports, as all data is exposed in the API, we can create them in the frontend.

roll:
> For diffing, I'm concerned that this approach would be difficult to change because the processors are on the lower levels of our architecture. Any changes here will propagate to the API and the Explorer.

Really I can't say more for now because you say right things about flexibility to have it on explorer but it breaks initial requirements to have this reports on API - https://github.com/opentrials/opentrials/issues/50 (as I read it). May be there is a plan to provide this data to other apps etc. I think we need more information here.

vitorbaptista:
Agreed. @pwalsh When you're around, could you clarify these requirements? Should the API provide the actual diffs?

pwalsh:
The API should provide the actual pre-calculated diffs IMHO.

roll:
Rebased and ready to merge. Requires https://github.com/opentrials/api/pull/42

roll:
CLOSED
for now in favor of https://github.com/opentrials/api/pull/55

PR #24, opened 2016-05-18
[#30] Rename problems to conditions

vitorbaptista:
opentrials/opentrials#30

I haven't touched the `health_conditions_or_problems_studied` field. Should we rename it? To what?

### DIFF
--- a/processors/actrn/extractors.py
+++ b/processors/actrn/extractors.py
@@ -66,10 +66,10 @@ def extract_trial(record):
     return trial
 
 
-def extract_problems(record):
+def extract_conditions(record):
     # TODO: record['health_conditions_or_problems_studied'] - free text some time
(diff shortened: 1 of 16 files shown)


roll:
@vitorbaptista 
As you're diving in so you know how to check things e.g. `make start nct`?

`health_conditions_or_problems_studied` - it's on `collectors` layer. So we're totally OK if some source have any field name.

roll:
:+1: 

(if you will remove `type` from table also `writers.write_condition` should be updated to stop passing `type` in database call)

vitorbaptista:
@roll I couldn't run because I didn't have a `WAREHOUSE_DB`. I'm not sure how to create one... Anyway, I removed `conditions.type` here as well.

PR #26, opened 2016-05-20
Feature/recruitment status enum

roll:
https://github.com/opentrials/opentrials/issues/98


roll:
@vitorbaptista 
please review

vitorbaptista:
@roll Two suggestions:
1. Trim and lowercase the the recruitment status, so the code becomes a bit more reliable for small changes;
2. Instead of normalizing as `statuses[record['foo']]`, do `statuses.get(record['foo'])`, so the code doesn't break if we couldn't find one of the record statuses. Obviously, this depends on what you prefer: ignoring a record we can't parse or moving on with incomplete data;

Related to point `2`, do we have a way to log those issues and monitor the crawling quality? It can be as simple as `if record['foo'] not in statuses: warn('Couldnt find %s' % record['foo'])`. This is a different issue, just wondering...

roll:
@vitorbaptista 
1. I didn't change initial values because it will be reviewed by domain experts. Also for situation when some value will fail and we need to find it in database for example. I see more cons than pros of slugifying it.
2. Good point. For now in `processors` I don't use `dict.get()` at all to make it fail on any unexpected values - it will be logged to our PaperTrail event logger.

vitorbaptista:
@roll I don't mean slugifying, just lowercasing and trimming. I agree with avoiding on being too clever, but I can't think of an example where just lowercasing/trimming would give us the wrong answer.

roll:
@vitorbaptista 
You're right about `ictrp`, others look good imo. Just thinking it will be easy to work with values synchronized to actual values in `warehouse` and registries.

roll:
@vitorbaptista 
fixed `ictrp` mapping

roll:
@vitorbaptista 
is it OK to merge?

vitorbaptista:
@roll Sure! :+1:

PR #27, opened 2016-05-20
[#111] Remove unused fields

vitorbaptista:
These were removed in opentrials/api@f6d087f730ab46fd17fa51cf65a39084486ded82

opentrials/opentrials#111

I wrote these changes by going to all `extractors.py` files and looking over what should be removed. Are there other places I should look as well?

While doing this, I was thinking that it would be useful to only specify fields that are used in each extractor, avoiding things like explicitly settings `'role': None`. That way, it's easier for us to change our DB later.


roll:
The most important change is `writers` - actual call to db.
Also it looks like you've forgot about some `role` and `context`.

---

> While doing this, I was thinking that it would be useful to only specify fields that are used in each extractor, avoiding things like explicitly settings 'role': None. That way, it's easier for us to change our DB later.

I've started with this approach writing like `intervention.get(data, {})` in `writers` but then found current way is more explicit (when `extractors` provide fully compliant to ideal model data). Let's think about it on the next iteration.

PS
For such type of things I use https://github.com/ggreer/the_silver_searcher

vitorbaptista:
@roll Went through the extractors again deleting the remaining `role` and `context` fields. Thanks for the hint on `the_silver_searcher`, didn't know it (I just use grep). Ideally, I would like a search tool that worked on Python's AST. [ASTSearch](http://takluyver.github.io/posts/astsearch-code-searching-that-knows-about-code.html) looks interesting, but haven't tried it yet.

About the implicit VS explicit, I do prefer to be as explicit as possible, and using defensive programming techniques (e.g. assert) to fail as soon as we get something unexpected. However, that should be used only for data we care about. In other words, my code shouldn't break if some value, that I don't care about in the first place, was changed. Specifically about these extra attributes, if they're not valid for all extractors, they should be optional in the DB. If that changes (and they become required), we should simply migrate the DB and fix whatever fails in our code. This makes our code as coupled as needed, but not more.

roll:
@vitorbaptista 
Yea I agree some changes could be added here. So explicit technique was important when all parts was moving so just to see the same things in every extractor was useful. But de-couple changes of `base` and `extractors` could be good already.

vitorbaptista:
@roll Please ping me when this PR is good to go :v:

roll:
@vitorbaptista 
also users of data from extractors - there is some things we're removing (again it uses here some strict approach for now with some duplication to have more simple code on early stages)
https://github.com/opentrials/processors/tree/master/processors/base/processors
https://github.com/opentrials/processors/tree/master/processors/base/writers

vitorbaptista:
@roll should be done now :+1:

roll:
:+1:

PR #28, opened 2016-05-23
Bug/empty-public-title

roll:
https://github.com/opentrials/opentrials/issues/155

Changes to be sure we write `public_title` for every trial.


roll:
@vitorbaptista 
please take a look on this small thing

processors/base/helpers.py
```diff
@@ -58,10 +58,24 @@ def default(self, obj):
...
+
+def get_optimal_title(*strings):
+    """Return first striped string with enough length or last striped string.
+    """
+    MIN_LENGTH = 10
+    for string in strings:
+        if not string:
+            continue
+        string = string.strip()
+        string = string.strip('.')
```
vitorbaptista: I'm not sure if this improves anything, but doing multiple strips like this might have some corner cases. For example, imagine that `str = 'foo bar.   '`, then `str.strip().strip('.')` would return `'foo bar'`, but if `str = 'foo bar  .'`, then the same code would return `'foo bar '`.

If this might be a problem, you can avoid this using: `str.strip('.' + string.whitespace)`. The `string.whitespace` is defined as `'\t\n\x0b\x0c\r '`, so you would strip all whitespace and all dots in a single pass, which always work.

This might be better or not, depending on what you expect to be in the string.

vitorbaptista:
@roll I added a comment. Regardless of it, this is good to go IMO :+1:

processors/base/helpers.py
roll: +1 `str.strip('.' + string.whitespace)` (initial thing was because some title are `....` but it's much better)

vitorbaptista: Depending on what's the use case here, `str.strip(string.punctuation + string.whitespace)` might be even better. This would remove any whitespace plus `'!"#$%&\'()*+,-./:;<=>?@[\\]^_`{|}~'`

roll: I'll check `string.punctuation` when will be reviewing updated database. I think strip things like `+` could be risky like `Trial of HIV+`. But may be subset of this chars.

PR #29, opened 2016-05-27
First version of the HRA collector filling for those fields that overlap with NCT

bzhr:
## Here's a simple version of the processor for those fields that overlap with NCT.

@roll @pwalsh 
- There's no data to add to locations/conditions/interventions.
- Some fields can be an empty string, do I need to check if it is empty string before insertion?

### DIFF
--- a/processors/hra/extractors.py
+++ b/processors/hra/extractors.py
@@ -1,0 +1,84 @@
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+from .. import base
+
+
+# Module API
+
+def extract_source(record):
+    source = {
+        'id': 'hra',
+        'name': 'Health Research Authority'
+    }
+    return source
+
+
+def extract_trial(record):
+
+    # Get identifiers
+    identifiers = base.helpers.clean_dict({
+        'hra_internal_identifier': record['application_id'],
+        'iras_project_identifier': record['iras_proj_id'],
+        'study_type_id': record['study_type_id']
+    })
+
+    # Get public title
+    public_title = base.helpers.get_optimal_title(
+        record['application_title'],
+        record['application_full_title'],
+        record['application_id']
+        )
+
+    trial = {
+        'primary_register': 'Health Research Authority',
+        'primary_id': record['application_id'],
+        'identifiers': identifiers,
+        'registration_date': record['publication_date'],
+        'public_title': public_title,
+        'brief_summary': record['research_summary'],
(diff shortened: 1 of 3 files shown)


processors/hra/extractors.py
```diff
@@ -0,0 +1,84 @@
...
+    return source
+
+
+def extract_trial(record):
+
+    # Get identifiers
+    identifiers = base.helpers.clean_dict({
+        'hra_internal_identifier': record['application_id'],
+        'iras_project_identifier': record['iras_proj_id'],
+        'study_type_id': record['study_type_id']
```
roll, line 26: We're interested here in `nct_id`, `isrctn_id` and `euctr_id` (eudra_ct). This HRA ids should be omitted for now.

processors/hra/extractors.py
```diff
@@ -0,0 +1,84 @@
...
+            {'name': record['establishment_org_post_code']}
+            ])
+    return organisations
+
+
+def extract_persons(record):
+    persons = []
+    persons.append({
+        'name': record['contact_name'],
+        'trial_id': record['application_id']
```
roll, line 82: Also here could go only `nct_id, isrctn_id and euctr_id`. Person without one of this id couldn't be added (current limitation).

roll:
@BenGosub 
I've added a few comments (we need other ids for deduplication system). I've checked code - it works on my dev environment.

processors/hra/extractors.py
bzhr, line 26: @roll these id's aren't present in the HRA API, there's `application_id` (hra internal id) and `iras_project_id` (IRAS project identifier).

roll, line 26: https://github.com/opentrials/collectors/pull/26/files#diff-ff5945e2fef10f894747606572ac5a8cR51

bzhr, line 26: I've checked this field it can take a value as: `00781976`, `NCT00672477`, `n.a`, `N/A`, `0`, `not_known`, should I check before insertion if the field is in the right format?

roll, line 26: Yes we need here ids in canonical form `NCT11111*`, `ISRCTN123*`, `EUCTR2011-003574-84-GB`. Ids we can't make compliant we don't need to add to `identifiers` list.

bzhr, line 26: If the entry is an 8 digit number, is it correct to add `NCT` in front of it?
E.g. `00781976` correct to `NCT00781976`?

roll, line 26: Yes. The same for `isrctn`. Please skip `euctr` (`eudra_ct`) (hra form doesn't fit with our current system).

PR #32, opened 2016-06-22
Bug/hra to publications

roll:
Moved HRA data to publications.


roll:
@vitorbaptista 
please take a look

processors/base/helpers.py
```diff
@@ -58,6 +58,16 @@ def default(self, obj):
...
 
 
+def clean_list(raw_list):
+    """Remove falsy values from list.
+    """
+    cleaned_list = []
+    for value in raw_list:
+        if value:
+            cleaned_list.append(value)
+    return cleaned_list
```
vitorbaptista, line 68: This is programming style and you're certainly aware, but just if you missed it, this could be rewritten as:

``` python
def clean_list(raw_list):
    return [value for value in raw_list
            if value]
```

Which seems more pythonic to me.

processors/hra/extractors.py
```diff
@@ -19,79 +19,43 @@ def extract_source(record):
...
+    abstract = record['research_summary'] or title
+    identifiers = base.helpers.clean_list({
+        _clean_identifier(record['nct_id'], prefix='NCT'),
+        _clean_identifier(record['euctr_id'], prefix='EUCTR'),
+        _clean_identifier(record['isrctn_id'], prefix='ISRCTN'),
     })
-    organisations.append({
-        'name': record['establishment_org'],
+    if identifiers:
+        abstract = '%s [%s]' % (abstract, '/'.join(identifiers))
```
vitorbaptista, line 38: Good idea to add the identifiers without changing the publications' schema :+1:

In the future, we should probably add those identifiers to the related trial itself.

processors/hra_linker/processor.py
```diff
@@ -0,0 +1,60 @@
...
+            else:
+                trial = {
+                    'primary_register': 'Health Research Authority',
+                    'primary_id': identifier,
+                    'identifiers': {
+                        'hra': identifier,
+                    },
+                    'public_title': identifier,
+                }
+                base.writers.write_trial(conn, trial, 'hra')
```
vitorbaptista: I imagine `base.writers.write_trial` will add the trials' `primary_source_id` to `hra`, right?

processors/base/helpers.py
roll, line 68: Yea here it could be used without code complication. Just because of my rule to do not use non obvious for all or control-flow complicating language concepts sometimes I miss opportunities like this.

processors/hra_linker/processor.py
roll: Correct)

vitorbaptista:
@roll If I understood correctly, the process for linking trials to HRA "publications" is:
1. For each HRA publication, do:
   1. For each HRA identifier, do:
      1. Look for a trial with that identifier
      2. If found, link the trial with the publication
      3. If not, create a new "empty" trial with just the identifier (similar to what we do with pubmed)

I'm not sure, but I'd think that it should be changed to:
1. For each HRA publication, do:
   1. Look for all trial with any of the identifiers we got from HRA;
   2. If found, link the trial with the publication and merge the HRA identifier list with the current trial identifier list (so we add any new identifiers from HRA);
   3. If not, create a new "empty" trial with all identifiers we found from HRA.

The reasoning behind this is that I think there's a 1-to-1 relationship between a research summary and a trial, so it doesn't make sense to use all identifiers as if each of them were relative to a different trial.

Considering our deadline (and that I'm not sure if my interpretation is correct), I'm happy to simply keep the code as is removing the creation of new empty trials and create a new issue about this so we can do it later.

What do you think?

roll:
@vitorbaptista 
Relationship is 1-to-many as minimum because there could be other research summaries sources. Also I'm not sure that one research summary could not be able to point to a few trials in some cases.

So I don't see a difference between research_summaries and publications for now. WDYT?

vitorbaptista:
Duh, you're right. The relationship is 1 trial for many summ) research summaries. However, are you sure that the other way is 1 to many as well? As in, a single research summary from HRA can point to many trials? 

If I remember correctly, there's only one identifier of each type (nct,...), right? If that's true, it seems to me as an indicator that the research summary only talks about a single trial. If that wasn't true, I'd expect it to be possible to link a research summary to multiple trial ids from the same registry.

roll:
We need an additional consultation about it when we will be adding new `research_summary` entity. 

But for now I think we need to use this approach (pubmed like) because it's anyway will be slightly re-written after new entity will be added.

PR #33, opened 2016-07-17
Feature/osf processor

### LINKED ISSUE #274: Make OSF integration
# Goal

Export data to OSF.

# Analysis

JamDB is a schema-less, immutable database that can optionally enforce a schema and stores provenance. It supports efficient full-text search, filtering by nested keys, and is accessible a REST API.

http://jamdb.readthedocs.io/en/latest/

We need to export our trials data to OSF using their API.

# Development

- [x] read and understand OSF documentation
- [x] write code for reading all trials with related entities from `database` (like api returns for `trials/<id>` endpoint)
- [x] write code for writing trials to OSF using batch requests
- [ ] export all trials to OSF

Estimated time: 4d

roll:
It fixes https://github.com/opentrials/opentrials/issues/274

### DIFF
--- a/processors/osf/processor.py
+++ b/processors/osf/processor.py
@@ -1,0 +1,14 @@
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import logging
+logger = logging.getLogger(__name__)
+
+
+# Module API
+
+def process(conf, conn):
+    pass
--- a/processors/osf/processor.py
+++ b/processors/osf/processor.py
@@ -4,11 +4,116 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import json
 import logging
+import requests
+import datetime
+from .. import base
 logger = logging.getLogger(__name__)
 
 
 # Module API
 
 def process(conf, conn):
-    pass
+
+    # Tables to export
+    TABLES = [
+        'locations',
+    ]
+
+    # Export tables
+    # http://jamdb.readthedocs.io/en/latest/index.html
+    token_ttl_seconds = 30*60 # api has limit 60*60 seconds
+    token_issued_time = datetime.datetime.now()
+    session = requests.Session()
+    for table in TABLES:
+
+        # Log started
+        logger.info('Started "%s" export', table)
+
+        # Ensure authentificated
+        if (datetime.datetime.now() - token_issued_time).seconds > token_ttl_seconds:
+            del session.headers['Authorization']
+        if 'Authorization' not in session.headers:
+            url = '%s/auth' % conf['OSF_URL']
+            res = session.post(url, json={
+              'data': {
+                'type': 'users',
+                'attributes': {
+                  'provider': 'osf',
+                  'access_token': conf['OSF_KEY'],
+                }
+              }
+            })
+            # Check status
+            # 200 - ok
+            if res.status_code not in [200]:
+                logger.error('Can\'t authentificate', table)
+                exit(1)
+            token = res.json()['data']['attributes']['token']
+            session.headers.update({'Authorization': token})
+            logger.info('Succesefully authentificated')
+
+        # Ensure collection exists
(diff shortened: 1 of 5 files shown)


roll:
@vitorbaptista please take a look (esp questions)

processors/osf/processor.py
```diff
@@ -0,0 +1,146 @@
...
+    # Export trials
+    # http://jamdb.readthedocs.io/en/latest/index.html
+    count = 0
+    limit = 100
+    token_ttl_seconds = 30*60  # api has limit 60*60 seconds
+    token_issued_time = datetime.datetime.now()
+    session = requests.Session()
+    for trials in _read_trial_groups(conn, limit=limit):
+
+        # Ensure authentificated
```
vitorbaptista: authentificated -> authenticated

processors/osf/processor.py
```diff
@@ -0,0 +1,146 @@
...
+                'type': 'users',
+                'attributes': {
+                  'provider': 'osf',
+                  'access_token': conf['OSF_KEY'],
+                }
+              }
+            })
+            # Check status
+            # 200 - ok
+            if res.status_code not in [200]:
```
vitorbaptista, line 47: Maybe change this to `res.status_code != 200`?

processors/osf/processor.py
vitorbaptista: There're a few others `authentificate`, so it'll be easier to do a search/replace

processors/osf/processor.py
```diff
@@ -0,0 +1,146 @@
...
+                logger.error('Can\'t authentificate')
+                exit(1)
+            token = res.json()['data']['attributes']['token']
+            session.headers.update({'Authorization': token})
+            logger.info('Succesefully authentificated')
```
vitorbaptista: Succesefully -> successfully

processors/osf/processor.py
```diff
@@ -0,0 +1,146 @@
...
+                'type': 'collections',
+                'attributes': {},
+            }
+        })
+        # Check status
+        # 201 - created
+        # 409 - conflict
+        if res.status_code == 201:
+            logger.info('Created collection "trials"')
+        elif res.status_code not in [201, 409]:
```
vitorbaptista: We already know here that `res.status_code != 201` (because of the previous if), so you can simply do `res.status_code != 409`. Also, why 409 is OK? Does it mean they already have the trial? It'd be good to explain this in a comment.

vitorbaptista:
@roll apart from a few typos and some minor stuff, this looks good to go IMHO.

As a programming style comment, I tend to prefer many small functions over a single big one. I think it improves the readability of the code. For example, in this PR, I would refactor to make it look like:

``` python
for trials in _read_trial_groups(conn, limit=limit):
    if _authentication_expired(token_issued_time):
        session.headers['Authorization'] = _authenticate(session)
    _ensure_collection_exists(conf, session)
    _export_trials(trials, conf, session)
```

With the current code snippets extracted to those functions. I feel it clarifies the code's intention and makes it easier to think in different levels of abstraction (i.e. I can see the actual functions' implementatino if needed, but can easily ignore them).

processors/osf/processor.py
roll, line 47: Normal flow here could be with many status codes so I use list by default to simplify changes.

roll:
About functions - I agree about good practice to make code modular but for tasks like this it could lead to increased cognitive complexity for reading/debagging than just plain code. It's because of many state things needed to pass to functions (and have in mind changing functions contexts) like counters, timers, session, trials etc. So this functions will be a really dirty not pure or even "half-pure".

vitorbaptista:
@roll Agreed. That's a sign that this function could be refactored as a class. That's not possible for now, given all the structure around processors, but might be something to keep in mind.

roll:
@vitorbaptista 
I think we shouldn't dive to deep here (or for similar processors) into any optimizations (performance or code style) because for now it looks like more like one-time thing (may be before next iteration but anyway).

A real game changer could be this one - https://github.com/opentrials/opentrials/issues/180 - if we will be having time for refactoring.

PR #34, opened 2016-07-18
First version of the code with a country normalization

### LINKED ISSUE #250: Normalize geographical location names
OpenTrials allows the user to search for trials in a specific geographical location (e.g. UK, Brazil, ...). However, the location names aren't normalized. This means that searching for "UK" or searching for "United Kingdom" might return different results. This also creates issues when we're trying to get statistics around trials per geographical location.

To solve this, we need to normalize the names into a canonical form (as a first step). I found a few tools that might help ([here](https://github.com/Corollarium/geograpy2), [here](https://github.com/elyase/geotext), [here](http://chairnerd.seatgeek.com/fuzzywuzzy-fuzzy-string-matching-in-python/), and [here](http://www.geonames.org/)). Many look abandoned, so I'm not sure which is best.
# Non Goals
- We won't deal with corner cases (e.g. [here](http://explorer.opentrials.net/trials/0cea0d2b-3d5b-4003-8372-b26b1f5f40f2), [here](http://explorer.opentrials.net/trials/fd8cf6b9-053e-463c-b1e9-04c59e470135), [here](http://explorer.opentrials.net/trials/d37f1364-b3fd-4a00-a24a-b7b599ba29e4) and [here](http://explorer.opentrials.net/trials/96c41ab1-abf7-4ed8-b435-cc9d3e5221c6)), just log them;
- We won't deal with synonyms (e.g. "USA" and "United States");
- We won't deal with geographical "levels" (i.e. a filter for "Europe" returning trials in "Germany");
# Expected Output

A helper method/function/class (whatever makes sense) in the [processor](https://github.com/opentrials/processors) that returns the canonical name of a geographical location. It should log all normalizations it's doing, including if it was successful or not. If it's unable to figure it out, it should return the same name it received.
# Acceptance Criteria
- [ ] Every location in OpenTrials appear only once (there're only [~300 at the moment](http://api.opentrials.net/v1/search/autocomplete/location), so it's easy to check)

Iararibeiro:
Some details about it, after testing all the suggested libs for @vitorbaptista  I decided use the one that @roll suggested combine with the csv table (is not a beautiful solution, but I gonna explain the decisions).
First the lib that I use resolve more problems than the ones suggested, it works if all the "normal" normalization cases and resolve some cases like synonyms ("US", 'USA" case). Another good point is because the lib offers a  search and results for more options, so if in the future you guys decide works with the ISO name, alpha or numeric that still will works.
I thought that I good option to deal with the corner cases or some problems that can show in the future was using some Fuzzy in the code, so if you can't find a name of the place using the lib, the code check in a table if exist a country that match in 80%. That works in cases like "United States of America" and common error like "Brasil", and in the future if you guys decided deal with the other corners cases in a different way is just add lines in that table. Again I thought that was a good improvement, and can be changed easily. 

It fixes https://github.com/opentrials/opentrials/issues/250


vitorbaptista:
Thanks @Iararibeiro, the code is looking good so far :+1:

Let me see if I understood what you've done. The process to normalize the name is:
1. Try getting the name using Python's iso3166
1.1. If that fails, return the country name closest to the one we're analyzing (using levenshtein distance, with at least 80% similarity)

Is that correct? Have you tested with the current locations in OpenTrials? How was the performance?

Iararibeiro:
yes @vitorbaptista, The have one more step.
1. Try getting the name using Python's iso3166
1.1. If that fails, return the country name closest to the one we're analyzing (using levenshtein distance, with at least 80% similarity)
1.2. If can't find any name with 80% of similarity, is returned the country_name received.

Make unit tests with the corner cases posted in the issue description and they works (I add some to test the similarity names), I execute the "make start nct" before and after the changes, I thought that the use of a table can make that slow but the checking isn't too common so the time was ok. And I also save both logs to compare and check, the log didn't show problems. If you guys have more tests to be execute I will like to do it.

vitorbaptista:
I've just taken the list of countries from the database. Could you check your code with it, so we can analyze its performance? I had to zip it to be able to upload to GitHub

[List of OpenTrials locations](https://github.com/opentrials/processors/files/369092/opentrials_locations.csv.zip)

processors/base/writers/location.py
```diff
@@ -70,3 +76,26 @@ def write_location(conn, location, source_id, trial_id=None):
...
+        # fuzzy logic to get the country name who get 80% of match rate.
+        with open(os.path.join(os.path.dirname(__file__),'countries.csv'), 'r') as f:
+            countries_data = csv.reader(f, delimiter= str(u','))
+            logger.debug('Location - %s: %s','not normalized', canonical_name)
+
+            for country in countries_data:
+                unicode_row = [x.decode('utf8') for x in country]
+                if (fuzz.ratio(location, unicode_row) > 80):
+                    canonical_name = countries.get(country[1]).name
+                    logger.debug('Location - %s: %s','not normalized with the most close result', canonical_name)
```
vitorbaptista: This `for country in countries_data` loop seems not to be doing what I think you'd want to. It goes over every country and, if it's `fuzz.ratio` if over `80`, it saves its name (and keeps going through the loop).

The problem here is that there can be countries that have the ratio, for example, `[60, 100, 85]`. If we're going over them in order, this code would return the name with ratio `85`, even though there's a perfect match in the dataset.

vitorbaptista: Also, does `fuzz.ratio` cleans the countries' names, like trimming whitespaces, lowercasing, etc.?

processors/base/writers/location.py
```diff
@@ -34,7 +40,7 @@ def write_location(conn, location, source_id, trial_id=None):
     timestamp = datetime.datetime.utcnow()
 
     # Get name
-    name = helpers.clean_string(location['name'])
+    name = helpers.clean_string(extract_CanonicalName(location['name']))
```
vitorbaptista: Instead of cleaning the output of `extract_CanonicalName`, it might be better to clean the input (i.e. `location['name']`). If we can find a match for the location, the result will already be cleaned (as it's the canonical name anyway). However, if we don't get a match, we'll return `location['name']` itself, which might need to be cleaned.

You can even do the cleaning inside the method itself, if that makes sense to you.

Iararibeiro: I understand your points @vitorbaptista and are good ones.
I didn't treat the case of having different ranges of similarity because 80% is a high number and I didn't saw any countries that gonna have so similar names. (80% means small number of letters like "of" or "s", but if you think can be a good improvement I can add that)
And I the fuzz.ratio do the lowercasing process.

vitorbaptista: I see. Still, if you think any match with >80% similarity is good, you could break the code after the first one is found (avoiding having to check all countries every time). Still, I'd guess returning the closest one (and not the last with >80% similarity) can give better results, even though I'd agree it should be quite rare to have close ties.

Are you sure that `fuzz.ratio()` lowercases/normalizes the names? I tried looking into their code, but couldn't find this normalization. Maybe I'm missing something, but it's better to confirm.

Iararibeiro:
Commit the changes suggested by @vitorbaptista and tested with the locations from open trials, the file helped a lot to understand what more need some work.
First, sorry Vitor you was right...the fuzzy doesn't do a lowercase and trim before compare, I don't know how I confuse that.
I used the file to test how the function works, some results (the file have 293 rows):
1. 77.16% of the rows are normalized (9.23% by similarity): In the majority of the cases the problems was errors in the country names (China as Chnia) or the alpha name like USA or DEU (Germany).
2. 22.86% of the rows are not normalized: In majority are corner cases, capitals or regions of countries (the last two can be worked in the future with a more complex solution in my opinion).

One error that happen: South America is normalized to South Africa, I thought that was weird but the two words have 87% of similarity. I test use the similarity level more high but the code loses cases of error like China as Chnia.

Iararibeiro:
The GitHub didnt attached the file, I forget that need put it in a zip.

[opentrials_locations.zip](https://github.com/opentrials/processors/files/371484/opentrials_locations.zip)

processors/base/writers/location.py
```diff
@@ -70,3 +76,27 @@ def write_location(conn, location, source_id, trial_id=None):
...
+            countries_data = csv.reader(f, delimiter= str(u','))
+            logger.debug('Location - %s: %s','not normalized', canonical_name)
+            best_similarity = 0
+            for country in countries_data:
+                unicode_row = [x.decode('utf8') for x in country]
+                country_name = unicode_row[0].lower().strip().replace(".", "")
+                if ((fuzz.ratio(location_name, country_name) >= 80) and 
+                    (fuzz.ratio(location_name, country_name) > best_similarity)):
+                    canonical_name = countries.get(unicode_row[3]).name
+                    logger.debug('Location - %s: %s','not normalized with the most close result', canonical_name)
```
vitorbaptista: The `best_similarity` is never updated. To avoid doing these two `fuzz.ratio` calls, you could simply set `best_similarity = 80` and then just test if `fuzz.ratio(...) >= best_similarity`, updating the `best_similarity` value.

vitorbaptista:
@Iararibeiro The code is looking good. So far, the main issues I can see are related on how to update the data files, both to keep it up to date and to deal with corner cases. We also need to deal with continents and at least have an idea on how to deal with cities.

Iararibeiro:
@vitorbaptista I fixed the bugs, sorry for that.
One suggestion is combine other lib who deals with regions and continents, so the comparing/checking can be in this order: continent > regions > countries > cities. 
Some of the corner cases like (over 40 countries, for example) I don't have any idea for a way do normalize it.

Iararibeiro:
I add some test cases, some of them have corner cases and will break. 
Make some changes in the code and now is checking the country capitals levels...That improve the number of normalized cases, I thought in add the continent check but the .csv file only have the code of continent not the names.
I search more about libs and some of them resolve one case or another, but doesnt resolve all the cases. So to resolve all problems and normalize all the cases, a good idea is combining libs.
The file with new results is attached.

And about the .csv used to extract the similar country names, one option to make it updated is run a script with some frequency that will compare if the file changed. I used the file suggested in the issue description:
https://github.com/datasets/country-codes/blob/master/data/country-codes.csv
I don't have a better idea about how refresh this file or adding cases, I will be very thankful if someone give suggestions.

[opentrials_locations.zip](https://github.com/opentrials/processors/files/377159/opentrials_locations.zip)

tests/test_location.py
```diff
@@ -0,0 +1,57 @@
+from processors.base.writers.location import normalize_location
+
+def test_simple_case(self):
+	#"""Testing the simple and easy normal case of normalization"""
```
roll: `#` is not needed here

roll:
It already looks as a big improvement. So now it's xxx locations instead of 294?

Hard-coding (mapping for not resolved ones) we was talking about? 

Also you could use https://github.com/frictionlessdata/tabulator-py (or `datapackage-py`) to read countries csv from web. There is no need to have a local file and even more script to manage it. For example you could create `COUNTRY_CODES` variable in `config.py` with data from `https://github.com/datasets/country-codes/blob/master/data/country-codes.csv` so it will be "actualized" on every processor start.

vitorbaptista:
- Remove magic numbers
- Use DictReader instead of the regular CSV reader to avoid magic numbers
- Extract `.lower().strip().replace(".", "")` into a lambda
- Fix the log messages: a location normalized by using the most similar name IS normalized (the current message says it isn't normalized)
- Log when we couldn't normalize at all
- Use test parametrization (http://docs.pytest.org/en/latest/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions)
- Make sure the coding style is consistent with the rest of the codebase (e.g. using underscores instead of CamelCase)

processors/base/writers/location.py
```diff
@@ -70,3 +76,37 @@ def write_location(conn, location, source_id, trial_id=None):
...
+def normalize_location(location):
+    location_name = location.lower().strip().replace(".", "")
+    try:
+        canonical_name = countries.get(location_name).name
+        logger.debug('Location - %s: %s normalized', location_name, canonical_name)
+    except KeyError:
+        canonical_name = location
+        # This part is for looking for matching cases (suggestion), uses the 
+        # fuzzy logic to get the country name who get 80% of match rate.
+        with open(os.path.join(os.path.dirname(__file__),'countries.csv'), 'r') as f:
```
vitorbaptista: Extract the CSV path to a variable and add a whitespace between `os.path.join` parameters

processors/base/writers/location.py
```diff
@@ -70,3 +76,37 @@ def write_location(conn, location, source_id, trial_id=None):
...
+                    logger.debug('Location - %s: %s not normalized, using the most similar name',
+                     location_name, canonical_name)
+                else:
+                    capital = unicode_row[21].lower().strip().replace(".", "")
+                    similarity = fuzz.ratio(location_name, capital)
+                    if similarity >= best_similarity:
+                        best_similarity = similarity
+                        canonical_name = countries.get(unicode_row[3]).name
+                        logger.debug('Location - %s: %s not normalized, using the most similar capital name',
+                        location_name, canonical_name)
```
vitorbaptista: Instead of using this loop, you can calculate all similarities at once doing something like:

``` python
_max_index_and_value = lambda arr: max(enumerate(arr), key=lambda x: x[1])
similarity_with_countries = [fuzz.ratio(location_name, country['name']) for country in countries]
max_index, max_value = _max_index_and_value(similarity_with_countries)
if max_value >= THRESHOLD:
    result = countries.get(countries[max_index]['ISO3166-1-Alpha-2']).name
else:
    # Try capitals
```

This could even be extracted to a separate function, as it's done for capitals and countries. It might help improve the readability, or maybe not. Try and see which you prefer.

vitorbaptista:
@roll I prefer to have the data needed by this code (i.e. `country-codes.csv`) in this repository, so we can tweak it as needed (for example, adding corner cases).

Iararibeiro:
@vitorbaptista I tried to used the DictReader as you suggested but the problem is do access a specific row, apparently is not possible without interacted over the structure. So one solution I founded is pass the csv to a tuple.

vitorbaptista:
To be fixed by https://github.com/opentrials/processors/pull/102

PR #36, opened 2016-07-27
Feature/contrib processor

### LINKED ISSUE #260: Add contributed data to our database
# Goal

We had a contribution of ~100 CSRs and some other documents. We need to get them all, upload individually to our S3 instance, and create the related entry in the `documents` table in the API database.

We should also document this process, as we'll be doing it more in the future.

This is blocked on #254

# Analysis

This documents are stored on s3 with link in `exporerdb`. Documents are in zip format. The process could be fully automated but the processor should includes specific to this contribution data like filename mapping etc. Initial implementation could be partially ad-hoc because we need to upload this documents ASAP.

# Development
- [x] create new `contrib_johnson` processor
- [x] add `explorerdb` credentials to env vars and `config.py`
- [x] add s3 credential to env vars and `config.py`
- [x] in code download archive and extract it 
- [x] in code write files metadata to `database.documents` table
- [x] create mapping between filenames in zip archive and metadata for `documents` table
- [x] in code upload files from zip to s3
- [ ] upload J&J documents

Estimated time: 2d

roll:
It fixes https://github.com/opentrials/opentrials/issues/260


roll:
@vitorbaptista 
https://github.com/opentrials/processors/blob/db89e85af51a15a06d18b0c6cef92a7a59de1aba/docs/processors/contrib.md

OK, before we don't have an UI for contribution uploading it could be workable.

processors/contrib/processor.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+# Internal
+
+def _download_documents(url, dirpath):
+    """Download documents to dirpath and extract if needed.
+    """
+    content = requests.get(url).content
+    # Archive of documents
+    if url.endswith('.zip'):
```
vitorbaptista: Use `url.lower().endswith('.zip')` to make iure it works even if the extension isn't lowercased

processors/contrib/processor.py
```diff
@@ -0,0 +1,139 @@
...
+            # Save document (if not exists)
+            table = conn['database']['documents']
+            if not table.find_one(trial_id=trial['id'], url=url):
+
+                # Upload document to S3
+                resource.Bucket(bucket).upload_file(filepath, key)
+
+                # Write document to database
+                table.insert({
+                    'id': uuid.uuid4().hex,
```
vitorbaptista: It won't make a difference here, but UUIDv4 wreaks havoc over indexes because their values, by design, are spread over all possible values. Because of this, it's better to use UUIDv1.

processors/contrib/processor.py
vitorbaptista: Better yet: use `zipfile.is_zipfile(content)`, which works even if the extension isn't `.zip`. You might have to `content.rewind()` afterwards, though.

vitorbaptista:
@roll I'm afraid we'll end up making hard to find mistakes when creating the regexps or defining the mapping between PDF and trials. Maybe using a processor wasn't a good idea after all.

I'm now thinking about a small Python script that you could execute locally. It could work like:

```
$ python ./opentrials_push_contributions.py ./data_contributions/
Looking for PDFs inside "./data_contributions"
Found PDF "foobar.pdf". Opening it.
Do you want to use this PDF? (y/n) y
What kind of document is it? (csr_report, epar_segment, ...) csr_report
What Trial is related to it? f9526184-1237-4393-847d-d0c2cb87da37
The file "foobar.pdf" is a "csr_report" and will be linked to trial "f9526184-1237-4393-847d-d0c2cb87da37". Is that correct? (y/n) y
Uploading and linking... Success! Created document with ID dafba7fc-5da0-44f0-829a-71eadfa70243
// ...
```

This will require you to have the credentials for S3 and the production DB locally, which is a big drawback. However, it feels less likely to cause mistakes, and shouldn't be hard to implement (we'll probably be able to reuse a lot of the code you already created).

What do you think?

roll:
The most obvious thing I think - such a discussion better to have after "In Analysis" step. 

I'm not a expected user of this script so of course I tends to follow the user expectations of workflow (even kinda belated=) but a few thoughts about it:

---

Using production credentials (any kind of access) at development machine is a real drawback as you said. It leads to things like https://github.com/opentrials/opentrials/issues/187 etc. So the best way to interact with production I suppose is `code->review->CI/CD->production`. E.g. I just don't have docker-cloud credentials locally so I just have no chance to break anything issuing some local commands.

---

For now workflow could be like this:
- update processor mapping
- push to master
- run processor on docker-cloud

```

INFO:processors.contrib.processor:Started extration from contrib "9e4f1280-41bf-11e6-8971-f99af8d5a820"
INFO:processors.contrib.processor:Document "nct00004941.pdf" saved with url "https://s3.eu-central-1.amazonaws.com/roll-opentrials-contrib/documents/a02fd141b8d06c4eaf2b42412ad408030d2a16ba.pdf"
INFO:processors.contrib.processor:Document "nct00034749.pdf" is already saved
WARNING:processors.contrib.processor:Document "JNJ-17335630-AAD Phase IIIb.pdf" is not in contrib mappings
WARNING:processors.contrib.processor:Document "nct00210860.pdf" has no matched trial
INFO:processors.contrib.processor:Finished extration from contrib "9e4f1280-41bf-11e6-8971-f99af8d5a820"
```

I don't feel really good about such a workflow (using Google Spreadsheet as mapping source could improve it) but I'm not sure why it's more error prone than manual processing of hundreds of files. My experience says opposite) Using mapping errors are not damaging - error goes to log and could be easily fixed. So wrong regexp just lead to file which will not be uploaded (should be uploaded on the next try).

---

Local script idea is also a reliable option esp if users could be non technical (we need to provide clear instruction) but I have a few concerns:
- on this step `What Trial is related to it? f9526184-1237-4393-847d-d0c2cb87da37` from where we've got trial_id? user input?
- if we'got contribution of 1000+ files - how much time it will take to process it file by file copying and reading trials identifiers?
- errors? human factor in an interactive routine work?

So for manual (small count of documents) uploading perfeclty will be to have a admin UI but I understand it's not a option now.

vitorbaptista:
@roll Good points. Considering that neither solution (script or the processor) is ideal, and that the processor is already done, let's merge it as is  :+1: (just see my comments on the code)

processors/contrib/processor.py
roll: Thanks, good point (`content.seek(0)` in python and it's needed but not documented in zip module though)

roll: Heh, not so easy `zipfile.is_zipfile` doesn't work with streams. Anyway `url.lower().endswith('.zip')` is good enough.

PR #38, opened 2016-08-18
FDA Approval History Processor

vitorbaptista:
The data from the `fda_dap` collector contains the approval history of FDA drugs. Each entry in the approval history can have many documents, and each document can be split in multiple PDFs.

For each entry in the `fda_dap` table, we:
1. Save the row's metadata in the `fda_approvals` table (if there's already an approval with the same ID, skip it);
2. For each document
   2.1 Download and merge the PDFs (if there're more than one)
   2.2 Upload to S3
   2.3 Upload to DocumentCloud
   2.4 Save to our DB

The way I structured the code is different from the other processors, which is something I'm not very keen on. However, I tried to make the code as clean as possible, and added a few (very few) tests. I couldn't add more tests because most of them would require access to a test database, which would mean configuring the testing infrastructure. This is something I want to do in the future, as our processors will only become more complex, but chosen not to do now.

I also tried to make the code idempotent. The biggest challenge in this was to use natural IDs, so the same ID is always generated for the same entity, and I also use the SHA-1 hash as key to upload to S3.

When you review this, please also review https://github.com/opentrials/collectors/pull/34 and https://github.com/opentrials/api/pull/68, which are its dependencies.

opentrials/opentrials#302

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -11,3 +11,7 @@ AWS_ACCESS_KEY_ID='<access_key>'  # optional
 AWS_SECRET_ACCESS_KEY='<secret_key>'  # optional
 AWS_S3_BUCKET='<bucket>'  # optional
 AWS_S3_REGION='<region>'  # optional
+AWS_S3_CUSTOM_DOMAIN='http://example.org'  # optional
+DOCUMENTCLOUD_USERNAME='<username>'  # optional
+DOCUMENTCLOUD_PASSWORD='<password>'  # optional
+DOCUMENTCLOUD_PROJECT='<project name>'  # optional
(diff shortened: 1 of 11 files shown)


roll:
@vitorbaptista 
Looks good AFAIK

Few comments:
- logging in `fda_dap` processor?
- `FDADAPProcessor` looks more like a set of functions than a real class

vitorbaptista:
@roll Added more logging messages to the FDA DAP processor. About if the `FDADAPProcessor` is a class or not, I'd argue that it is, considering it has data + methods that use them. I could've simply created a few unbounded functions, however most of them would need the `conn` and `conf` variables, so their signatures would become `def foo(conn, conf, ...)`. This, IMHO, is a big indication that they're actually a class in disguise.

roll:
:+1: 

About `conf` and `conn` - just FYI - I've added it for high-level testability to do not use some hacks and patchs in tests. So processors (and collectors) partially pure functions (not using environment at all, but of course having side effects):

``` python
def test_some_processor():
    conf = {}  # use needed
    conn = {}  # use test databases
    processor(conf, conn)
```

Example:
- https://github.com/opentrials/collectors/blob/master/tests/test_hra.py#L17-L21
- https://github.com/opentrials/collectors/blob/master/tests/conftest.py#L30-L40

vitorbaptista:
@roll That dependency injection will be very helpful :+1:

PR #39, opened 2016-08-21
Updated deduplication system and bug fixes

### LINKED ISSUE #240: Improved deduplication system
# Description

Based on current experience of data deduplication I could suggest some implementation improvements and strategies to use. I'll write it as tasks for the reading convenience but it's of course for a discussion (cc @vitorbaptista @pwalsh @benmeg).

Ideas highlights:
- use only identifiers for automatic matching
- improve querying duplicates implementation (facts system)
- use curators for matching records based on pre-processed similarity

# Tasks
- [ ] instead of `trials.facts` use `records.identifiers` (or indexable analogue) to search for the same trial records. Current implementation based on "stateful" `trials.facts` field should be replaced because of maintainability problems (e.g. after hra removing facts should be cleaned from hra strings etc).
- [ ] create processor and database table for calculation trials similarity. We could start with scientific titles match then follow with probabilistic and machine learning methods. As output table like `trial_id_1, trials_id_2, method, similarity, is_marked_as_matching, updated_by_user_id` (not real naming just to show the idea). Could be split for 2 tables etc.
- [ ] create processor for reading similarity table to dynamically merge/unmerge records based on user marked matches
- [ ] create some redirects table because trial pages could be merged at any time (saving history of dedup process is a thing to think about)

---
- [ ] move UI tasks to different issue after discussion
- [ ] create UI for similarity table - as similar records on trial page with ability to mark it as duplicate
- [ ] create UI for similarity table - as a list of similar record pairs ordered by similarity to match by users/curators. So some hired curators or enthusiasts could highly effective processing this list from top to bottom.

### LINKED ISSUE #270: trial appears to be linked to multiple records without matching name or ID
http://explorer.opentrials.net/trials/869d45c8-66b0-4720-8507-270142d6ca84/discrepancies

This trial appears to be linked to numerous records that don't share its name or any ID numbers. 

It would be super helpful if we could trace why a record was tagged as a particular trial - thanks!

roll:
# Overview

The main change is updated deduplications system. Previous version was based on `trials.facts` field which was a problem because this field was like a moving part (always growing) of the system and was almost not suitable for things like unlinking records from trial.

Updated deduplication system works like this:
- we process new record from `warehouse`
- find another record with intersection in `records.identifiers`
- get `records.trial_id` from found record
- find trial using this `trial_id` (so this new record will be linked to this trial)

Other improvements:
- cleaned `processors.base` - less entities like `readers`
- updated data model following to https://github.com/opentrials/api/pull/69
- rebased on master id for euctr (without country code) filling `records.identifiers` (because there is only one `euctr` key)
- fixed `records.identifiers` extraction for euctr (for https://github.com/opentrials/opentrials/issues/270)

Next steps:
- implement `dedup` processor which will be linking/unlinking existent trials based on their records (additional system to have records properly linked after source changes)

It partially:
- fixes https://github.com/opentrials/opentrials/issues/240
- fixes https://github.com/opentrials/opentrials/issues/270


vitorbaptista:
@roll Could you check the merge conflicts here?

PR #41, opened 2016-08-25
[#267] Parse trial statuses into "status" and "recruitment_status"

vitorbaptista:
Depends on https://github.com/opentrials/api/pull/71

This gives us more granularity to know the status of a trial. The mapping of the statuses comes from https://github.com/opentrials/opentrials/issues/267#issuecomment-239802828.

opentrials/opentrials#267


roll:
In writers (record, trial) we use explicit approach so status also should be added there.

Also do you think manage this mappings as one merged in helpers is a bad idea? This could simplify new extractors writing. I'm going to move identifiers check inside `write_trial` via helper. The same could be for statuses.

roll:
In writers (record, trial) we use explicit approach so status also should be added there.

Also do you think manage this mappings as one merged in helpers is a bad idea?

vitorbaptista:
@roll You mean moving this logic to `processors/base/writers` instead of it being in each separate processor? If so, I don't think it's a good move, as it breaks encapsulation. The identifiers might make sense, as it's the same for any processor (right?).

I also added the `status` to the writers. Nice catch.

PR #45, opened 2016-08-26
Feature/identifiers check

roll:
# Overview

Added a helper to check identifiers before adding to the system.


processors/base/helpers.py
```diff
@@ -69,14 +59,39 @@ def clean_list(raw_list):
...
+        'kct': r'^KCT',
+        'nct': r'^NCT',
+        'ntr': r'^NTR',
+        'pactr': r'^PACTR',
+        'per': r'^PER',
+        'rbr': r'^RBR',
+        'rpcec': r'^RPCEC',
+        'takeda': r'^TAKEDA',
+        'tctr': r'^TCTR',
+        'who': r'^U',
```
vitorbaptista: It would be good to have more strict regexps where possible (e.g. `^NCT\d+`)

vitorbaptista:
I added a comment about being more strict with the regexps, specially because there're quite a few very open (e.g. `^U`). It would be good to expand at least the easy ones (e.g. NCT), but I'm happy to handle this later.

processors/base/helpers.py
roll: I see it's just a start of ongoing process of discovering it. Many identifiers has strict structure like `EUCTR-dddd-dddddd` etc. Esp after `identifiers` table introduction I suppose we make identifiers management very important part of the system.

But this starter covers all errors we had for now (I'll re-check after processors finish).

vitorbaptista: Unless I'm missing something, it should be quite easy to extract the identifiers' patterns from our current database.

roll: Not so easy) Because there could be many special cases of thinks like zeros in NCT And then we will start to add secondary ids like industry ones (https://github.com/opentrials/opentrials/issues/329), DOI etc. So I just wanted to say - there could be I suppose some amount of logic working only on identifiers. Area concentrating many domain knowledge and requiring some analysis step by step.

vitorbaptista: Still, simple things like `r'^NCT\d+$'` is quite safe and already an improvement IMHO. We can leave the difficult ones for later, but there're many low-hanging fruits here. I'm happy to do this later, though.

roll: I think I will be able to extract some tomorrow before merging.

PR #46, opened 2016-08-30
Feature/record linker

roll:
# Overview

PR purpose (along with other PRs for this issues):
- https://github.com/opentrials/opentrials/issues/240
- https://github.com/opentrials/opentrials/issues/270
- https://github.com/opentrials/opentrials/issues/310
- https://github.com/opentrials/opentrials/issues/295
- https://github.com/opentrials/opentrials/issues/292


roll:
@vitorbaptista 
please take a look on logic - it should finalize many problems fixing (dedup update) but I'm not feel very well with current record-trial approach. I'll create an issue for a future with my thoughts. But for now we need to fix this dedup problems and this should work. Just needed double check and proper db backup :smiley:

processors/base/helpers.py
```diff
@@ -202,9 +202,12 @@ def find_trial_by_identifiers(conn, identifiers):
...
         records = list(conn['database'].query(query))
-        if not records:
-            continue
-        trial = conn['database']['trials'].find_one(id=records[0]['trial_id'].hex)
+        for record in records:
+            if not record['trial_id']:
+                continue
+            trial = conn['database']['trials'].find_one(id=record['trial_id'].hex)
+            if trial:
+                break
```
vitorbaptista: The query could be simplified to:

`SELECT records.trial_id, identifier.key, identifier.value FROM records, json_each_text(records.identifiers::json) AS identifier WHERE identifier.value IN ('foo', 'bar')`

This doesn't take in consideration the identifiers' source. To do so, we could change the query to something like:

`SELECT records.trial_id FROM records WHERE trial_id IS NOT NULL AND (identifiers @> '{"nct": "NCT00322725"}' OR identifiers @> '{"nct": "FOO"}')`

``` python
base_query = 'SELECT trial_id FROM records WHERE trial_id IS NOT NULL AND ({where_clause})'
where_clause = ' OR '.join(['identifiers @> ?'] * len(identifiers.keys()))
query = base_query.format(where_clause=where_clause)

identifiers_clause = [json.dumps({source: identifier})
                      for source, identifier in identifiers.items()]

records = conn['database'].query(query, identifiers_clause)
if records:
    # No need to test if the "trial_id" exists, as we're protected by
    # Postgres' foreign key constraints
    return conn['database']['trials'].find_one(id=records[0]['trial_id'].hex)
```

This reduces the number of queries we need to do to 2, regardless on how many identifiers a trial has. We could even get the trial in a single query, but that might be overkill.

processors/record_linker/processor.py
```diff
@@ -0,0 +1,72 @@
...
+        if len(trial_records) == 1:
+            continue
+        linked_identifiers = set()
+        for trial_record in trial_records:
+            if record['id'] == uuid.UUID(trial_record['id']).hex:
+                continue
+            linked_identifiers.update(trial_record['identifiers'].items())
+        if not linked_identifiers.intersection(record['identifiers'].items()):
+            conn['database']['records'].update(
+                {'id': record['id'], 'trial_id': None}, keys=['id'])
```
roll: NOTE: just delete record?

processors/base/helpers.py
roll: Thanks. I'll try it. 

But now I'm trying to focus only on business logic (like more queries but easier to understand what code does). "premature optimization is the root of all evil (or at least most of it) in programming"(c)?)

roll:
Also better to split this into two processors with proper names I suppose.

processors/record_linker/processor.py
```diff
@@ -0,0 +1,72 @@
...
+        'trials_locations',
+        'trials_organisations',
+        'trials_persons',
+        'trials_publications',
+    ]
+
+    # Unlink wrong linked records
+    # [will be reprocessed on next processors run]
+    count = 0
+    for record in base.helpers.iter_rows(conn, 'database', 'records', orderby='id'):
```
vitorbaptista: As far as I can see, there's no need for `orderby='id'` here. Also, I haven't tested it, but you could maybe add `conn['database']['records'].columns.trial_id != None` to remove the records without `trial_id`.

processors/record_linker/processor.py
```diff
@@ -0,0 +1,72 @@
...
+        # Already unlinked
+        if not record['trial_id']:
+            continue
+        trial_records = list(conn['database']['records'].find(trial_id=record['trial_id']))
```
vitorbaptista: I haven't tested, but maybe it's possible to add `conn['database']['records'].columns.id != record['id']` as another param to the `find()` call, so you won't need to bother removing the current record from the array.

processors/record_linker/processor.py
```diff
@@ -0,0 +1,72 @@
...
+        # Record not linked to others
+        if len(trial_records) == 1:
+            continue
+        linked_identifiers = set()
+        for trial_record in trial_records:
+            if record['id'] == uuid.UUID(trial_record['id']).hex:
```
vitorbaptista: I'm curious: why `record['id'] == trial_record['id']` doesn't work? They both come from the same column type...

processors/record_linker/processor.py
```diff
@@ -0,0 +1,72 @@
...
+                continue
+            linked_identifiers.update(trial_record['identifiers'].items())
+        if not linked_identifiers.intersection(record['identifiers'].items()):
+            conn['database']['records'].update(
```
vitorbaptista: As we talked, actually removing the records is better

processors/record_linker/processor.py
roll: do you remember our elasticsearch bug before walkthrough? :smiley:

processors/record_linker/processor.py
roll: Because this `uuid` type in database is a nightmare. I was debugging it 30 min. Sometimes it's like `adsdfasdfds` sometimes like `sadfsda-sdafa-sdf` sometimes uuid object needing hex call. I will never be using it again))) For uuid it could be plain text in db..

processors/record_linker/processor.py
```diff
@@ -0,0 +1,72 @@
...
+from .. import base
+logger = logging.getLogger(__name__)
+
+
+# Module API
+
+def process(conf, conn):
+
+    # Trial m2m
+    TRIAL_TABLES = [
```
vitorbaptista: Added a new PR that makes this easier: https://github.com/opentrials/api/pull/73

After that's merged (hopefully before this), you won't need to bother with those relationship tables, but will still have to delete from the `documents` table.

roll: I think better to get an error on PaperTrials if there are documents related to bad trial. So then we could decide what to do with it.

processors/record_linker/processor.py
```diff
@@ -0,0 +1,72 @@
...
+                {'id': record['id'], 'trial_id': None}, keys=['id'])
+            logger.info('Unlinked record: %s', record['identifiers'])
+            count += 1
+        if count and not count % 100:
+            logger.info('Unlinked %s wrong linked records', count)
+    logger.info('Unlinked %s wrong linked records', count)
+
+    # Remove trials without records
+    count = 0
+    for trial in base.helpers.iter_rows(conn, 'database', 'trials', orderby='id'):
```
vitorbaptista: No need for `orderby` here, and also I think you can add `conn['database']['trials'].columns.source_id != 'pubmed'`, so could remove lines 55 and 56.

Actually, considering you're looking for trials without records, I would do an actual query like:

`SELECT trials.id FROM trials LEFT JOIN records ON records.trial_id = trials.id WHERE records.trial_id IS NULL AND trials.source_id != 'pubmed'`

This makes this code much quicker, as there's only a single request needed and we're only getting the trial id, and not every column. On the other hand, it's loading all IDs in memory instead of iterating over them. This shouldn't be a problem, though, as considering that one UUID has only 16 bytes, even a million UUIDs will require only 16 MB of memory (disconsidering possible Python/Dataset overheads).

PR #47, opened 2016-08-31
Improved deduplication

roll:
Improved:
- `record_remover` to remove records have identifier segments
- `find_trial_by_identifiers` to force deduplication first

### DIFF
--- a/processors/base/helpers.py
+++ b/processors/base/helpers.py
@@ -186,12 +186,13 @@ def iter_rows(conn, dataset, table, orderby, bufsize=100, **filter):
             yield row
 
 
-def find_trial_by_identifiers(conn, identifiers):
+def find_trial_by_identifiers(conn, identifiers, ignore_record_id=None):
     """Find first trial matched by one of passed identifiers.
 
     Args:
         conn (dict): connection dict
         identifiers (dict): identifiers dict (nct: <id>, euct: <id>, ...)
+        ignore_record_id (str): skip record with this id (for better dedup)
 
     Returns:
         dict: trial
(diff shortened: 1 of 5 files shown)


processors/base/helpers.py
```diff
@@ -203,9 +204,12 @@ def find_trial_by_identifiers(conn, identifiers):
...
         query = QUERY % json.dumps({source: identifier})
         records = list(conn['database'].query(query))
-        if not records:
-            continue
-        trial = conn['database']['trials'].find_one(id=records[0]['trial_id'].hex)
+        for record in records:
+            if (ignore_record_id and record['id'].hex == uuid.UUID(ignore_record_id).hex):
+                continue
+            trial = conn['database']['trials'].find_one(id=record['trial_id'].hex)
+            if trial:
```
vitorbaptista, line 211: As far as I can see, the trial will always exist, guaranteed by the foreign key constraints on `records` table.

processors/record_remover/processor.py
```diff
@@ -13,23 +12,50 @@
...
+        # Sort segments and get the biggest segment
+        segments = list(sorted(segments, key=lambda s: len(s), reverse=True))
+        biggest_segment = segments[0]
+
+        # Delete all records without intersection with the biggest segment
+        for record in records:
+            idset = set(record['identifiers'].items())
+            if not biggest_segment.intersection(idset):
+                if conn['database']['records'].find_one(id=record['id']):
+                    conn['database']['records'].delete(id=record['id'])
```
vitorbaptista, line 55: You just got that `record` from the DB, so it's safe to delete it without testing that it exists. At most, we should handle the exception if it doesn't exist.

processors/base/helpers.py
roll: Good catch, this function a good candidate for a refactoring (just don't have time for it now at all, added you new query as comment). But we should be very careful here because it's core of all dedup system.

processors/base/writers/trial.py
```diff
@@ -30,8 +30,11 @@ def write_trial(conn, trial, source_id):
     create = False
     timestamp = datetime.datetime.utcnow()
 
-    # Get trial object
-    object = helpers.find_trial_by_identifiers(conn, identifiers=trial['identifiers'])
+    # Get trial object (first try to ignore source record for better dedup)
+    object = helpers.find_trial_by_identifiers(conn, trial['identifiers'],
```
vitorbaptista, line 34: There should be no bad side effects here, but reusing reserved words like this can create some tricky bugs.

PR #48, opened 2016-08-31
Fixes to fda data

vitorbaptista:
This is part of the fixes needed for opentrials/opentrials#361. I also changed the URL we're uploading documents to in S3 to be in folder "./documents/fda", so it's easier to find/delete/filter the documents on S3 when needed.


roll:
LGTM

PR #49, opened 2016-09-01
Run tests before building container

vitorbaptista:
The tests should run much faster than the container building, so we don't build
containers when the code is broken.


victor-nitu:
Oh, such a small one, I overlooked it 😅 

All good! 👍

PR #50, opened 2016-09-02
[#289] Create results Document if trial has results

vitorbaptista:
The same for Records, so in the end a trial will have all documents referenced
by its primary registry and all other sources as well.

We're not able to determine from which record a document comes from, as the
"documents" table only has a relationship to the Trial. However, we know their
source, so if there's only one record from that source, we found our culprit.
Otherwise we can only reduce the number of possibilities.

Maybe we'll need to add a "record_id" to the documents in the future, but not
now.

opentrials/opentrials#289


victor-nitu:
All good here, manually tested most of it, tests are fine, I have nothing to add but this: 👍

PR #52, opened 2016-09-02
Log and ignore exceptions on FDA_DAP

vitorbaptista:
This allow us to continue with the next record, even if one fails for some
reason. Catching "Exception" usually is a bad practice, but it's the correct
approach in this case.

@nightsh I'm merging this before you can review so we can keep the processor running over the weekend. Please check it afterwards.


victor-nitu:
👍 😁

PR #55, opened 2016-09-12
Remove unknown DocumentCloud documents

vitorbaptista:
This processor removes all documents from DocumentCloud that we have no
reference of in our database ("files" table).


docker-cloud.yml
```diff
@@ -288,3 +288,15 @@ trial-remover:
     LOGGING_URL:
   tags:
     - opentrials
+
+remove-unknown-documentcloud-docs:
+  image: okibot/processors
```
victor-nitu, line 293: I totally overlooked the fact that we have a bot building docker images from this repo. Neat!

victor-nitu:
👍

PR #56, opened 2016-09-12
[#279] Processor to create (new) PyBossa tasks for new or modified indications

victor-nitu:
This new processor makes a query for new or modified indications, extracted from FDA approvals. The result is used to create new PyBossa tasks, deduplicate and `POST` them to the server as new tasks.

The main issue here is with API rate limiting, so there are a few mechanisms to avoid data loss due to 429 status codes from the API. Basically, when rate limit is reached, the processor waits a bit before sending out new payloads.


processors/pybossa_indications/processor.py
```diff
@@ -0,0 +1,127 @@
...
+import os
+import time
+import requests
+import pbclient as pbc
+import logging
+logger = logging.getLogger(__name__)
+
+PYBOSSA_URL = os.environ['PYBOSSA_URL']
+PYBOSSA_API_KEY = os.environ['PYBOSSA_API_KEY']
+project_id = 4228
```
vitorbaptista: This should be an env variable as well

victor-nitu: What if we have multiple projects? Should I create something like `PYBOSSA_PROJECT_INDICATIONS` and follow this convention for subsequent ones?

processors/pybossa_indications/processor.py
```diff
@@ -0,0 +1,126 @@
...
+
+    submit_tasks(tasks)
+
+
+def pybossa_rate_limitation(endpoint):
+    # This should be called before actual requests to avoid getting HTTP 429s
+    res = requests.get('{}/api/task'.format(PYBOSSA_URL))
+    if int(res.headers['X-RateLimit-Remaining']) < 30:
+        logger.warn('Rate limit reached, will sleep for 5 minutes')
+        time.sleep(300)  # Sleep for 2 minutes
```
vitorbaptista: The log message has the wrong time

processors/pybossa_indications/processor.py
```diff
@@ -0,0 +1,126 @@
...
+                'name': doc['name'],
+                'url': doc['url']
+            })
+        tasks.append(task)
+    logger.debug('{} tasks in the database'.format(len(tasks)))
+
+    submit_tasks(tasks)
+
+
+def pybossa_rate_limitation(endpoint):
```
vitorbaptista: Parameter is unused

processors/pybossa_indications/processor.py
```diff
@@ -0,0 +1,126 @@
...
+    rows = list(conn['database'].query(query % (bufsize, offset)))
+    if not rows:
+        logger.info('No new and modified indications found for tasks creation')
+        return
+    create_tasks(rows)
+
+
+def create_tasks(rows):
+    tasks = []
+    for key, group in groupby(rows, lambda x: x['id']):
```
vitorbaptista: When you're not using some variable, name it `_`. In this case, it would be:

``` python
for _, group in groupby(rows, lambda x: x['id']):
```

Also, if possible, try naming the variables for what they contain. For example, instead of `group`, use `documents` or similar (if they have documents)

processors/pybossa_indications/processor.py
```diff
@@ -0,0 +1,126 @@
...
+        for doc in group:
+            task['supplement_number'] = str(doc['supplement_number'])
+            task['approval_type'] = str(doc['type'])
+            task['approval_id'] = str(doc['id'])
+            task['meta_id'] = str(doc['meta_id'])
+            task['notes'] = doc['notes'] or ""
+            task['documents'].append({
+                'name': doc['name'],
+                'url': doc['url']
+            })
```
vitorbaptista: You can change this for:

``` python
task = {
  'supplement_number': str(doc[0]['supplement_number']),
  # ...
}

task['documents'] = [{'name': doc['name'], 'url': doc['url']}
                     for doc in group]

tasks.append(task)
```

Also, instead of `doc['notes'] or ""` use `doc.get('notes', '')`

Do you need to cast everything to `str()`? Just asking because all fields except `supplement_number` and maybe `id` should be strings already.

processors/pybossa_indications/processor.py
vitorbaptista: Rename it to something like `sleep_if_reached_rate_limit`

processors/pybossa_indications/processor.py
```diff
@@ -0,0 +1,126 @@
...
+        logger.warn('Rate limit reached, will sleep for 5 minutes')
+        time.sleep(300)  # Sleep for 2 minutes
+        return
+
+
+def get_existing_ids(PROJECT_ID):
+    existing_ids = []
+    tasks = []
+    offset = 0
+    limit = 100
```
vitorbaptista: Not sure how big are the tasks, but 100 seems like a very small limit.

processors/pybossa_indications/processor.py
victor-nitu: I had issues with those two while working on some other aspects, so I did it for everything. Think it's safe to cast just those two?

As for the list comprehension, yes, that looks exactly like the idea I didn't have. Thanks! 👍

processors/pybossa_indications/processor.py
```diff
@@ -0,0 +1,126 @@
...
+
+    # Make sure we have enough requests left
+    pybossa_rate_limitation('task')
+    while True:
+        collection = pbc.get_tasks(PROJECT_ID, limit=limit, offset=offset)
+        if len(collection) > 0:
+            tasks.extend(collection)
+            offset += limit
+        else:
+            break
```
vitorbaptista: This could be rewritten as:

``` python
while True:
    pybossa_rate_limitation('task')
    last_id = tasks[-1] and tasks[-1].id
    collection = pbc.get_tasks(PROJECT_ID, limit=limit, last_id=last_id)
    tasks.extend(collection)
    if not collection:
        break
```

This removes the need for an offset (which should make PyBossa's query faster, see https://github.com/PyBossa/pybossa-client#on-queries-and-performance), moves the rate limits to inside the `while` loop, and reorganizes the code a bit. I'm not sure if `tasks[-1].id` exists, but there should be a similar call if not.

processors/pybossa_indications/processor.py
```diff
@@ -0,0 +1,126 @@
...
+    for task in tasks:
+        existing_ids.append(task.info['meta_id'])
+    return existing_ids
```
vitorbaptista: `return [task.info['meta_id'] for task in tasks]`

processors/pybossa_indications/processor.py
```diff
@@ -0,0 +1,126 @@
...
+
+def submit_tasks(tasks):
+
+    # Set up PyBossa connection
+    pbc.set('endpoint', PYBOSSA_URL)
+    pbc.set('api_key', PYBOSSA_API_KEY)
+
+    # Get existing IDs
+    existing_ids = get_existing_ids(PROJECT_ID)
+    if not isinstance(existing_ids, list):
```
vitorbaptista: `get_existing_ids` returns either a `list` or throws an exception (if `pbc.get_tasks()` has thrown one), so this check shouldn't be needed. You can use `if existing_ids` if you'd like to know if there're any IDs.

processors/pybossa_indications/processor.py
victor-nitu: Every time we call `pybossa_rate_limitation()` we make an additional request. That's why I tried to reduce its usage.
As for `last_id`, I think `pybossa-client` has it built in. Will check.

processors/pybossa_indications/processor.py
victor-nitu: Ah nice, I thought this goes against readability so I didn't do much of it :)

processors/pybossa_indications/processor.py
victor-nitu: That's the maximum payload we can get, I'm afraid...

processors/pybossa_indications/processor.py
victor-nitu: 👍 sounds reasonable

docker-cloud.yml
```diff
@@ -253,6 +253,18 @@ pubmed-linker:
...
 
+pybossa-indications:
+  image: okibot/processors
+  command: make start pybossa_indications
+  # restart: always
+  environment:
+    WAREHOUSE_URL:
+    DATABASE_URL:
+    EXPLORERDB_URL:
+    LOGGING_URL:
```
vitorbaptista: You also need `PYBOSSA_URL` and others.

processors/pybossa_indications/processor.py
```diff
@@ -0,0 +1,118 @@
...
+import os
+import time
+import requests
+import pbclient as pbc
+import logging
+logger = logging.getLogger(__name__)
+
+PYBOSSA_URL = os.environ['PYBOSSA_URL']
+PYBOSSA_API_KEY = os.environ['PYBOSSA_API_KEY']
+PROJECT_ID = os.environ['PYBOSSA_PROJECT_INDICATIONS']
```
vitorbaptista: Instead of hardcoding them here, add those on https://github.com/opentrials/processors/blob/master/processors/base/config.py and get this from the `conf` dict passed to `process()`

processors/pybossa_indications/processor.py
```diff
@@ -0,0 +1,118 @@
...
+    for key, group in groupby(rows, lambda x: x['id']):
+        group = list(group)
+        first = group[0]
+        print(str(first['meta_id']))
+        task = {
+            'supplement_number': str(first['supplement_number']),
+            'approval_type': first['type'],
+            'approval_id': str(first['id']),
+            'meta_id': str(first['meta_id']),
+            'notes': first.get('notes', '')
```
vitorbaptista: Do we use things like `supplement_number`, `approval_type` and others in the interface? If not, I would rather not add them here (and don't get them from the SQL query). As far as I can see, the `fda_approval_id` should be enough.

Also, if you're using a specific ID (e.g. `fda_approval_id`), name it as such and not a general `meta_id`, as it serves as a form of documentation on what ID it's about.

victor-nitu: Well, I thought you had something in mind with them when writing the initial query, so I left them there and forgot to ask you back about their importance. Great that you spotted them.

As for the `meta_id`, I fully agree. I sincerely can't remember when and why I came up with it.

Will update in minutes.

processors/pybossa_indications/processor.py
```diff
@@ -0,0 +1,107 @@
...
+    else:
+        logger.info('No new and modified indications found for tasks creation')
+
+
+def _create_tasks(conf, rows):
+    tasks = []
+    for key, group in groupby(rows, lambda x: x['fda_approval_id']):
+        group = list(group)
+        first = group[0]
+        print(str(first['fda_approval_id']))
```
vitorbaptista: Use `logger` instead of `print` (if this message is needed)

processors/base/config.py
```diff
@@ -47,6 +47,12 @@
 DOCUMENTCLOUD_PASSWORD = os.environ.get('DOCUMENTCLOUD_PASSWORD')
 DOCUMENTCLOUD_PROJECT = os.environ.get('DOCUMENTCLOUD_PROJECT')
 
+# PyBossa
+
+PYBOSSA_URL = os.environ.get('PYBOSSA_URL')
+PYBOSSA_API_KEY = os.environ.get('PYBOSSA_API_KEY')
+PROJECT_ID = os.environ.get('PYBOSSA_PROJECT_INDICATIONS')
```
vitorbaptista: Change to `PYBOSSA_PROJECT_INDICATIONS` (and not a general `PROJECT_ID`)

vitorbaptista (changes requested):

processors/base/config.py
victor-nitu: I totally agree. Funny though, I actually thought I did it.

processors/pybossa_indications/processor.py
victor-nitu: It isn't needed. It escaped, thanks for spotting it.

processors/pybossa_indications/processor.py
```diff
@@ -0,0 +1,107 @@
...
+    # Get stuff from the DB
+    query = """
+    SELECT documents.id AS document_id,
+        fda_approvals.id AS fda_approval_id,
+        documents.name,
+        files.url
+    FROM documents
+    INNER JOIN fda_approvals ON documents.fda_approval_id = fda_approvals.id
+    INNER JOIN files ON documents.file_id = files.id
+    WHERE documents.url IS NOT NULL
```
victor-nitu: Also this has to be changed (missed it).

victor-nitu: Done.

processors/base/config.py
victor-nitu: Done.

processors/pybossa_indications/processor.py
victor-nitu: Done.

processors/pybossa_indications/processor.py
```diff
@@ -0,0 +1,106 @@
...
+        files.url
+    FROM documents
+    INNER JOIN fda_approvals ON documents.fda_approval_id = fda_approvals.id
+    INNER JOIN files ON documents.file_id = files.id
+    WHERE files.url IS NOT NULL
+    AND fda_approvals.type = 'New or Modified Indication'
+    GROUP BY fda_approvals.id,
+            documents.id,
+            files.id
+    ORDER BY fda_approvals.id
```
vitorbaptista: Indent the string

vitorbaptista (approved):

processors/pybossa_indications/processor.py
vitorbaptista: Don't worry with it, just did it myself to avoid you having to context switch.

vitorbaptista:
@nightsh Your code was looking fine to me. However, I started looking deeper into it to create the new PyBossa processor for our deduplication project and got a bit carried away. In hindsight, I should've merged your code and then started working on it, but haven't.

I'm assigning it back to you so you can review my changes. I changed the way we check for the rate limits to use only the status code of the response we get (i.e. not doing another request to check the limits). The biggest change was extracting the PyBossa-related code to a separate class (to be reused by our other processors) and adding the logic to keep the tasks in our database in sync with the tasks in PyBossa, adding/removing/updating them as needed.

processors/base/helpers/pybossa_tasks_updater.py
```diff
@@ -0,0 +1,113 @@
...
+        get_keys = lambda values: [get_key(value) for value in values] # noqa: E731
+
+        existing_tasks = self._get_existing_tasks()
+        existing_tasks_info = [task.data['info'] for task in existing_tasks]
+        existing_ids = get_keys(existing_tasks_info)
+        new_ids = get_keys(tasks_data)
+
+        tasks_to_add = [task_data for task_data in tasks_data
+                        if get_key(task_data) not in existing_ids]
+        tasks_to_delete = [existing_tasks[i]
```
victor-nitu, line 45: That's an inspired addition to the processor. Good thinking!

processors/base/helpers/pybossa_tasks_updater.py
```diff
@@ -0,0 +1,113 @@
...
+                           for i, task_info in enumerate(existing_tasks_info)
+                           if get_key(task_info) not in new_ids]
+
+        tasks_to_update = []
```
victor-nitu, line 49: Are we allowed to update existing tasks? Even if they already have received answers?

victor-nitu:
@vitorbaptista it's a very good addition that further shapes up the desired modularity for our PyBossa related processors. Following this, I would probably rewrite a bit of the indications processor, to make it use more of the logic in the tasks updater, but this will probably be a good candidate for either the hack day or a later update.

Bottom line, I'm quite happy with it. TBH, I didn't run it locally, but by going through all the code I'd say it's merge material, feel free to do it 👍

processors/base/helpers/pybossa_tasks_updater.py
vitorbaptista: Good question. I haven't tested with tasks with answers, but I think the correct behaviour here would be to remove the existing answer and redo the task from the beginning. Hopefully PyBossa does it :pray:

PR #57, opened 2016-09-12
[#388] Sync `files.text` with the OCRd text from DocumentCloud

vitorbaptista:
This processor updates the column `files.text` in our database with the
contents of the full text extraction from DocumentCloud.

Fixes opentrials/opentrials#388


processors/sync_text_from_documentcloud/processor.py
```diff
@@ -0,0 +1,56 @@
...
+def _get_text(client, doc_id):
+    doc = None
+    text = None
+
+    try:
+        doc = client.documents.get(doc_id)
+    except Exception as e:
+        is_http_error_403 = (getattr(e, 'code', None) == 403)
+        if is_http_error_403:
+            raise e
```
victor-nitu, line 38: Any idea why is this happening?

processors/sync_text_from_documentcloud/processor.py
```diff
@@ -0,0 +1,56 @@
...
+        logger.debug(msg)
+
+    return text
+
+
+def _documentcloud_client(conf):
+    username = conf['DOCUMENTCLOUD_USERNAME']
+    password = conf['DOCUMENTCLOUD_PASSWORD']
+
+    return documentcloud.DocumentCloud(username, password)
```
victor-nitu, line 56: ouch, the alternative import looks even more redundant :D 
 `from documentcloud import DocumentCloud as documentcloud`

processors/sync_text_from_documentcloud/processor.py
```diff
@@ -0,0 +1,56 @@
...
+import documentcloud
+import logging
+from .. import base
+logger = logging.getLogger(__name__)
+
+
+def process(conf, conn):
+    query = (
+        'SELECT id, documentcloud_id FROM files'
+        ' WHERE documentcloud_id IS NOT NULL'
```
victor-nitu: I'd go for a triple quoted multiline string on this one, just for the ease of maintainability. But it really doesn't matter that much in this case, up to you.

victor-nitu:
All seems to be good, ping me if you feel like I should run it. If not, I'd merge it. If yes, you need to share some `.env` bits :)

processors/sync_text_from_documentcloud/processor.py
vitorbaptista: It happens when the password is wrong or you don't have permissions. It shouldn't happen normally, catching here just to, if it does, we stop processing altogether.

PR #58, opened 2016-09-13
[#389] Overwrite trials without records from the same source

vitorbaptista:
@roll assigning this to you because you have the context. Please let me know if you're unable to review.

We used to deduplicate trials/records based on their public titles. That
created too many false positives, so we changed the code to only match on
identifiers for now. However, this left some bogus data in our database.
Consider the following:

There's a trial FOO which was matched on a trial from NCT based on its title.
This means that we have in our database a Trial with `source_id == 'nct'`, its
corresponding `record`, and some other `records` from other sources (e.g.
EUCTR) matched by title.

After we changed our deduplication process to use only identifiers, our
processors went over the database removing all records that didn't match on
identifiers. This meant that the corresponding NCT records didn't match
anymore and was removed. The EUCTR record is kept, as it matches on ID.

Now we have a Trial with `source_id = 'nct'` with different `records`, none of
which is from NCT. When we run the EUCTR processor, it'll try to overwrite this
trial, but won't be able to because NCT has a higher priority than EUCTR. Our
database will then be in an invalid state.

The long-term solution for this is to revamp the way we handle identifiers,
extracting them to a separate table so we can remove all trials and recreate
them with the same IDs as needed. We can't do that now otherwise we would loose
the current OpenTrials IDs.

For the short-term, we simply check if a trial has at least one record from
that same source. If not, we overwrite it, regardless of the source priorities.

Fixes opentrials/opentrials#389


roll:
:+1:

PR #59, opened 2016-09-22
Cochrane processor

georgiana-b:
Processor for extracting risk of bias results from Cochrane reviews.


georgiana-b:
Using the fuzzy search of http://www.ncbi.nlm.nih.gov to get PubMed IDs seems a bit of a hassle so  I ran the processor with different ways of matching (risk of bias results to trials) to see it's worth keeping it. Here are the results:
- 73 matches using the fuzzy search of http://www.ncbi.nlm.nih.
- 93 matches using OpenTrials API  fuzzy search - leads to overmatches if we search only by `title` so it can be a viable solution only if we add  `authors` attribute to the publications in `trials` index or make another index for publications
- 53 matches without any fuzzy search

processors/base/writers/risk_of_bias.py
```diff
@@ -0,0 +1,138 @@
...
+
+    """
+    create = False
+    timestamp = datetime.datetime.utcnow()
+
+    # Read object
+    query_params = {'dataset': 'database', 'table': 'risk_of_biases',
+                    'orderby': 'updated_at', 'source_url': rob['source_url'],
+                    'study_id': rob['study_id']}
+    existing = [r for r in helpers.iter_rows(conn, **query_params)]
```
vitorbaptista: From https://github.com/opentrials/api/pull/87, there can't be multiple ROB with the same `source_url` and `study_id`, so you can simplify this query by doing `conn['database']['risk_of_biases'].find_one(study_id=rob['study_id'], source_url=rob['source_url'])`

processors/base/writers/risk_of_bias.py
```diff
@@ -0,0 +1,138 @@
...
+    if existing:
+        object = existing[-1]
+    else:
+        object = None
```
vitorbaptista: I know this pattern is used in other parts of the code, but don't use Python's reserved words as variables (we'll remove the others). You can do:

``` python
rob = conn['database']['risk_of_biases'].find_one(study_id=rob['study_id'], source_url=rob['source_url'])

if not rob:
    rob = {
        # ...
    }
```

processors/base/writers/risk_of_bias.py
```diff
@@ -0,0 +1,138 @@
...
+
+    # Create object
+    if not object:
+        object = {}
+        object['id'] = uuid.uuid4().hex
```
vitorbaptista: Never use `uuid4` for ID fields. Use `uuid.uuid1().hex` instead.

processors/base/writers/risk_of_bias.py
```diff
@@ -0,0 +1,138 @@
...
+    timestamp = datetime.datetime.utcnow()
+
+    # Read object
+    query_params = {'dataset': 'database', 'table': 'risk_of_bias_criterias',
+                    'orderby': 'updated_at', 'name': rob_criteria['name']}
+    existing = [rob_crt for rob_crt in helpers.iter_rows(conn, **query_params)]
+    if existing:
+        object = existing[-1]
+    else:
+        object = None
```
vitorbaptista: Same as my comments on `write_rob()`:
- Don't use Python's reserved words as variable names
- Prefer using `rob_criteria = conn['database']['risk_of_biases'].find_one(name=rob_criteria['name'])`

processors/base/writers/risk_of_bias.py
```diff
@@ -0,0 +1,138 @@
...
+
+    # Create object
+    if not object:
+        object = {}
+        object['id'] = uuid.uuid4().hex
+        object['created_at'] = timestamp
+        object['name'] = rob_criteria['name']
```
vitorbaptista: Instead of declaring the variable as `{}` and then changing its attributes, use:

``` python
rob_criteria = {
    'id': uuid.uuid1().hex,  # Note we're using UUID1 instead of UUID4
    # ...
}
```

processors/base/writers/risk_of_bias.py
```diff
@@ -0,0 +1,138 @@
...
+        create = True
+
+    # Update object
+    object.update({
+        'updated_at': timestamp
+    })
```
vitorbaptista: As you're updating only a single attribute, prefer `rob_criteria['updated_at'] = timestamp`.

processors/base/writers/risk_of_bias.py
```diff
@@ -0,0 +1,138 @@
...
+
+    # Log debug
+    logger.debug('Risk of bias - %s: %s',
+        'created' if create else 'updated', rob['source_url'])
+
+    return object['id']
+
+
+def write_rob_criteria(conn, rob_criteria):
+    """Write risk_of_bias_criteria to database.
```
vitorbaptista, line 65: As a `risk_of_bias_criteria` only has a `name` attribute, and we use the `name` to find it in the database, updating doesn't make much sense (as we're updating a row without changing any of its values). This method should only write the `risk_of_bias_criteria` if it doesn't exist, or return its `id` if it does.

processors/base/writers/risk_of_bias.py
```diff
@@ -0,0 +1,138 @@
...
+
+    # Log debug
+    logger.debug('Risk of bias criteria- %s: %s',
+        'created' if create else 'updated', rob_criteria['name'])
+
+    return object['id']
+
+
+def write_rob_rob_criteria(conn, rob_rob_criteria):
+    """Write risk_of_bias-risk_of_bias_criteria to database.
```
vitorbaptista, line 104: The `docstrings` are incorrect, as you're always returning `None`. I understand that you did so because it has a composite primary key. Maybe instead of having a `write_rob_rob_criteria()` function, it would be better to write it as part of the `write_rob_criteria()` function.

processors/cochrane/extractors.py
```diff
@@ -0,0 +1,74 @@
...
+                    '3.1': 'performance bias',
+                    '3.2': 'detection bias',
+                    '4': 'incomplete outcome',
+                    '5': 'selective reporting',
+                    '6': 'other bias'}
+    results = []
+
+    for result in review_results:
+        rob_name = result.get('rob_name', '')
+        matches = {id: val for id, val in correlations.items()
```
vitorbaptista: `id` is a reserved word in Python. Don't use it as a variable name. I usually use `key` instead.

processors/cochrane/extractors.py
```diff
@@ -0,0 +1,74 @@
...
+                    '3': 'blinding',
+                    '3.1': 'performance bias',
+                    '3.2': 'detection bias',
+                    '4': 'incomplete outcome',
+                    '5': 'selective reporting',
+                    '6': 'other bias'}
+    results = []
+
+    for result in review_results:
+        rob_name = result.get('rob_name', '')
```
vitorbaptista: As a small defensive programming technique, you can use `rob_name = result.get('rob_name') or ''` instead, to cover the case where `result['rob_name'] == None`

processors/cochrane/extractors.py
```diff
@@ -0,0 +1,74 @@
...
+        matches = {id: val for id, val in correlations.items()
+                   if val in rob_name.lower()}
+        matched_keys = sorted(list(matches.keys()))
+        if len(matched_keys) > 1:
```
vitorbaptista: You can simplify this by defining `id = None` before this `if`, as this'll allow you to remove the two `else: id = None`

processors/cochrane/processor.py
```diff
@@ -0,0 +1,152 @@
...
+            matches = conn['database']['trials_publications'].find(publication_id=pub['id'])
+            trial_ids.extend([match['trial_id'] for match in matches])
+
+    return trial_ids
+
+
+def match_by_pubmed_id(conn, pubmed_id):
+    """Find correspondent trial based on pubmed_id"""
+
+# NOTE: Match directly to `database` publications after `pubmed_id` is added
```
vitorbaptista: Use `FIXME` instead and fix the indentation

processors/cochrane/processor.py
```diff
@@ -0,0 +1,152 @@
...
+
+    return trial_ids
+
+
+def match_by_nct_id(conn, nct_id):
+    """Find correspondent trial based on nct_id"""
+
+    query_criteria = {'nct': 'NCT' + nct_id}
+    trial = base.helpers.find_trial_by_identifiers(conn, query_criteria)
+    return [trial['id']]
```
vitorbaptista: `trial` will be None if it can't be found, so you need to test for this case here, just changing the last line (100) to:

``` python
if trial:
    return [trial['id']]
```

processors/cochrane/processor.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+def extract_ref_identifiers(reference):
+    """Extract identifiers from reference"""
+
+    reference_ids = {}
+    useful_ids = {'NCT': 'nct_id', 'PUBMED': 'pubmed_id', 'PMID': 'pubmed_id'}
+    for id in useful_ids:
```
vitorbaptista: Don't use `id` as a variable name. You could use something like `_id`.

processors/cochrane/processor.py
```diff
@@ -0,0 +1,152 @@
...
+                if len(matching_trials) == 0:
+                    matching_trials.extend(match_by_reference(conn, reference))
+
+                if len(set(matching_trials)) == 1:
+                    trial_id = matching_trials[0]
+                    extractors = base.helpers.get_variables(
+                        extractors_module, lambda x: x.startswith('extract_'))
+                    base.processors.process_risk_of_biases(conn, extractors,
+                                                           latest_rev, trial_id)
+                    break
```
vitorbaptista: Log if there're more than 1 matched trial.

tests/processors/cochrane/test_extractors.py
```diff
@@ -0,0 +1,26 @@
...
+            {'result': 'YES', 'rob_name': 'Incomplete outcome data'},
+            {'result': 'YES', 'rob_name': 'Reporting bias'},
+            {'result': 'NO', 'rob_name': 'Free of other bias'}
+        ]
+        processed_results = extractors.extract_review_results(results)
+        expected = [{'name':'1', 'value':'unknown'},
+                    {'name':'2', 'value':'no'},
+                    {'name':'3.2', 'value':'yes'},
+                    {'name':'4', 'value':'yes'},
+                    {'name':'6', 'value':'no'}]
```
vitorbaptista: Instead of using IDs as names, use the actual human-readable names

vitorbaptista (changes requested):

processors/base/writers/risk_of_bias.py
```diff
@@ -0,0 +1,126 @@
...
+        ex_rob = {
+            'id': uuid.uuid1().hex,
+            'created_at': timestamp,
+            'source_url': rob['source_url'],
+            'study_id': rob['study_id'],
+        }
+        create = True
+
+    # Update object
+    ex_rob.update({
```
vitorbaptista: Could  check here if `ex_rob.get('trial_id') != rob['trial_id'] or ex_rob.get('source_id') != rob['source_id']`, so only update if needed

processors/base/writers/risk_of_bias.py
```diff
@@ -0,0 +1,126 @@
...
+            'created_at': timestamp,
+            'name': rob_criteria['name'],
+        }
+        create = True
+
+    ex_rob_crt['updated_at'] = timestamp
+
+    # Write object
+    conn['database']['risk_of_bias_criterias'].upsert(ex_rob_crt, ['id'],
+                                                      ensure=False)
```
vitorbaptista: Here we're only creating the `rob_crt`. Updating it is useless, as we just change the `updated_at` timestamp.

processors/cochrane/extractors.py
```diff
@@ -0,0 +1,73 @@
...
+            results.append({'name': name, 'value': result['result'].lower()})
+
+    return results
+
+
+def extract_source(record):
+    return {
+        'id': 'cochrane',
+        'name': 'Cochrane Library',
+        'type': 'other',
```
vitorbaptista: I think this should be "Cochrane Schizophrenia". Also add `url: http://schizophrenia.cochrane.org/`

processors/cochrane/processor.py
```diff
@@ -0,0 +1,163 @@
...
+                matching_trials = []
+                for id_name, id_value in identifiers.items():
+                    found_trials = match_by_identifier(conn, id_name, id_value)
+                    matching_trials.extend(found_trials)
+
+                if len(matching_trials) == 0:
+                    matching_trials.extend(match_by_reference(conn, reference))
+
+                unique_trails = set(matching_trials)
+                if len(unique_trails) == 1:
```
vitorbaptista: Typo `trails` -> `trials`

processors/cochrane/processor.py
```diff
@@ -0,0 +1,163 @@
...
+                    trial_id = matching_trials[0]
+                    extractors = base.helpers.get_variables(
+                        extractors_module, lambda x: x.startswith('extract_'))
+                    base.processors.process_risk_of_biases(conn, extractors,
+                                                           latest_rev, trial_id)
+                    break
+                elif len(unique_trails) > 1:
+                    logger.debug('Several matching trials found for reference %s. Matched trials ids: %s.',
+                                 reference, matching_trials)
```
vitorbaptista: Use `unique_trials` instead

processors/cochrane/processor.py
```diff
@@ -0,0 +1,163 @@
...
+
+def extract_ref_identifiers(reference):
+    """Extract identifiers from reference"""
+
+    reference_ids = {}
+    useful_ids = {'NCT': 'nct_id', 'PUBMED': 'pubmed_id', 'PMID': 'pubmed_id'}
+    for _id in useful_ids:
+        for value in reference.values():
+            try:
+                match = re.search(r'{0}[\s]*[^\d]*(\d+)'.format(_id), value).group(1)
```
vitorbaptista: This regexp matches things like `NCT lorem ipsum foo bar 12345678`. Also, extract the regexp to a variable, as you're using the same on line 130.

processors/base/writers/risk_of_bias.py
georgiana-b: Isn't that what `updated_at` is for? :slightly_smiling_face:

georgiana-b:
@vitorbaptista  Sorry for  the bulk of commits. I pulled with `rebase` from master. The latest changes that haven't been reviewed so far are in the last 3 commits (de27f0e, 58290cb, 04fcc18 ).

vitorbaptista (approved):

PR #60, opened 2016-09-26
[WIP] [#396] New processor to merge trial and record identifiers

victor-nitu:


victor-nitu:
Some issues I'm having so far:

First run, the processor correctly counts and updates everything. Second run, it finds some rows **from the previous result set** that still need updating. Third pass, same thing, but from the 2nd result set.

It's like it silently fails a bunch of updates. No exceptions are raised.

On a snapshot of the API DB dating `2016-09-08 20:14:38.000000000 +0300`, I have found **53473** records to update, of which **960** were still updated on the second pass and **12** on the third.

I even tried the desperate method of calling the update twice in the processor, "just in case". Same.

Questions:
1. Am I missing something obvious? Something less obvious?
2. Is it reproducible in any other environments than mine?
3. Is it a processor issue or a library issue? (dataset / sqlachemy)
4. Any suggestions?

vitorbaptista:
Merged https://github.com/opentrials/processors/pull/61 instead of this.

PR #62, opened 2016-09-29
[#415] HRA records / `source_url` that can be used by humans

victor-nitu:


processors/base/writers/publication.py
```diff
@@ -43,7 +43,7 @@ def write_publication(conn, publication, source_id):
         create = True
 
     # Write object only for high priority source
-    if create or source_id in ['pubmed']:
+    if create or source_id in ['pubmed', 'hra']:
```
victor-nitu: @vitorbaptista @roll 
Why?
_(I added `'hra'` in that list, but feels like a hack that we might not need anymore)_

vitorbaptista: Agreed. We should remove that.

victor-nitu: @vitorbaptista I will assign you to this PR, can you do the modification? I'm not sure if I should remove all the conditions or just the `source_id` ones 😅

PR #63, opened 2016-09-30
[#424] Allow updating publications

vitorbaptista:
This has the bad side effect of always updating the `updated_at` time, even if
we don't change the original data. We'll have to handle that afterwards.

Fixes opentrials/opentrials#424


vitorbaptista:
Merging this to fix an issue in production :+1:

victor-nitu:
Maybe we should open a new issue for... the new issue? I'd be happy to tackle it (as well as the hardcoded list I can't find right now, the one that made me question my sanity when updating HRA records) after we're done with the `Launch` milestone.

vitorbaptista:
Done opentrials/opentrials#467

PR #64, opened 2016-09-30
[WIP] [#398] Add trial deduplication tasks to PyBossa

vitorbaptista:
We're comparing trials two by two. To understand how we create the tasks,
consider a database with trials A, B, C and D. The tasks created will be (A, B),
(B, C) and (C, D). This won't test all possible cases, because they are in the
millions with our current database. This is just an initial pass. With this
logic, we'll create NUMBER_OF_TRIALS - 1 tasks.

There's a challenge here on how to upload this to CrowdCrafting, as it only
allows 300 requests per 15 minutes. In that speed, it'll take more than 10 days
to add the ~330k tasks. Not to say that 330k tasks is already a lot.

We'll need to filter out more trials to make it feasible, specially considering
that deciding if two trials are the same isn't a trivial task.

opentrials/opentrials#398


georgiana-b:
I'm wondering why we don't create groups of trials potentially the same and compare only trials within the groups with each other. I created [an example that creates groups based on same `public_title`](https://query.opentrials.net/queries/154). Then we could compare the trials withing each group of `trial_ids`. For example [the trials from the first group](https://query.opentrials.net/queries/155) do seem very similar.

This relies on the presumption that some fields are more important indicators of similarity than others. In the query I wrote: 

```sql
SELECT * FROM
  (SELECT t.rank, array_agg(t.id) AS trial_ids FROM
    (SELECT id, rank() OVER (ORDER BY public_title) FROM trials) AS t 
  GROUP BY t.rank
  ) AS r
WHERE array_length(r.trial_ids, 1) > 1 ORDER BY r.rank ASC;
```
we can reduce the number of groups even more by refining the query given to `OVER`.
What do you think?

vitorbaptista:
This is a good approach, but I think it'll still create a big number of trials. For instance, let's assume we have 600k trials in total and 300k deduplicated. This means we have, on average, 2 records on each trial. 2 records per trial and 300k trials means we'll have 300k tasks.

Say that we add title and reduce the number of trials to 150k (lots of trials with the same title), so on average we'll have 4 records per trial. There're 6 possible combinations of 2 elements in a group of 4 records. So we'll have 6 combinations * 150k groups: 900k tasks.

We must filter the possibilities even more, but that'll only be possible when we improve our deduplication process to use crowdsourcing only on the harder cases.

vitorbaptista:
Closing this for now. See https://github.com/opentrials/opentrials/issues/398#issuecomment-281632352 for reasoning.

PR #66, opened 2016-10-01
[#429] Fix document writer

vitorbaptista:
They don't have a `documentcloud_url` anymore, and the uniqueness constraint is
with (trial_id, url, file_id, fda_approval_id).

opentrials/opentrials#429


victor-nitu:
👍

PR #67, opened 2016-10-04
Handle whitespaces in identifiers

georgiana-b:
As requested in opentrials/opentrials#368.


processors/base/helpers/__init__.py
```diff
@@ -86,6 +86,10 @@ def clean_identifiers(identifiers):
     }
     result = {}
     for key, value in identifiers.items():
+        try:
+            value = re.sub(r'\s', '', value)
+        except TypeError:
+            pass
```
vitorbaptista, line 96: I like the idea, but unfortunately there're a few identifiers with whitespace (e.g. http://www.takedaclinicaltrials.com/browse/summary/AD-4833_409%20(PHARMO)?protocol_id=#overview), so we can't remove them all.

processors/base/helpers/__init__.py
```diff
@@ -128,7 +132,7 @@ def find_list_of_identifiers(text):
 
     # Pattern could be improved based on a extended
     # clinical trial identifiers format analysis
-    PATTERN = r'(%s[\d-]{3,})'
+    PATTERN = r'(%s[\s]*[\d-]{3,})'
```
vitorbaptista: Minor minor thing that I wouldn't bother with usually, but as you'll remove the code from https://github.com/opentrials/processors/pull/67/files#r81792086, you can also simplify this a bit. There's no need to add the `[]` in `[\s]`, as it's only a single character (it's needed in `[\d-]` because there we want either `\d` or `-`), so you can rewrite this as `r'(%s\s*[\d-]{3,})'`

processors/base/helpers/__init__.py
georgiana-b: Some identifiers with spaces will not pass this function because they don't comply with [the regex rules](https://github.com/georgiana-b/processors/blob/c7cfb2aabf229620e2f14e31cc696f2c73276be0/processors/base/helpers/__init__.py#L67-L86). I wrote this because even after I changed the regex in `find_list_of_identifiers`, ISRCTN and NCT ids with spaces were excluded here. Is there a way to know which identifiers are allowed to have spaces and where?

vitorbaptista: This is kinda hacky, but you could then change to something like `re.sub(r'^([^\s+])\s+', r'\g<1>', value)` instead. This removes the whitespaces between the `NCT` and whatever comes afterwards. To allow us to debug any issue, you can use:

``` python
new_value, num_changes = re.subn(r'^([^\s+])\s+', r'\g<1>', value)
if num_changes:
    logger.debug('Removed whitespaces from identifier "%s" to "%s"', value, new_value)
    value = new_value
```

processors/base/helpers/__init__.py
```diff
@@ -132,7 +136,7 @@ def find_list_of_identifiers(text):
 
     # Pattern could be improved based on a extended
     # clinical trial identifiers format analysis
-    PATTERN = r'(%s[\s]*[\d-]{3,})'
+    PATTERN = r'(%s\s*[\w\d-]{3,}(?:\s*\([\w]*\))?)'
```
georgiana-b: This has been changed because the previous one didn't match identifiers like `TAKEDA-AD-4833_409 (PHARMO)` at all.

vitorbaptista:
@georgiana-b I changed your regexp to one more strict. It won't handle identifiers like `AD-4833_409 (PHARMO)`, which is why you wrote that regexp in the first place, but it should keep the number of false positives a bit lower.

My concern with that is that I was concerned we would increase our number of false positives, as we don't have good regexps to validate identifiers yet. Whenever we improve them, we can reuse the regexp you wrote.

PR #68, opened 2016-10-05
[#453] Use NCT's actual enrollment size if there's no anticipated one

vitorbaptista:
Sometimes we have just one of the two.

Fixes opentrials/opentrials#453


victor-nitu:
let's merge it! 👍

PR #69, opened 2016-10-05
[#432] Validate urls before writing to database

georgiana-b:
opentrials/opentrials#432


tests/processors/base/writers/test_document.py
```diff
@@ -0,0 +1,46 @@
...
+        conn = _get_mock_conn()
+        conn['database']['documents'].find_one.return_value = None
+        document = mock.MagicMock(spec_set=dict)
+        document.__getitem__.return_value = 'mockingbird'
+        document.get.return_value = 'mockingbird'
+        helpers_mock.validate_remote_url.return_value = False
+        written_document = writers.write_document(conn, document)
+        helpers_mock.validate_remote_url.assert_called_with('mockingbird')
+
+        assert written_document == None
```
vitorbaptista: Tests with none should use `foo is None`, not `foo == None` (see http://stackoverflow.com/questions/23086383/how-to-test-nonetype-in-python/23086405#23086405 for reasoning)

tests/processors/base/writers/test_source.py
```diff
@@ -0,0 +1,46 @@
...
+        conn = _get_mock_conn()
+        conn['database']['sources'].find_one.return_value = None
+        source = mock.MagicMock(spec_set=dict)
+        source.__getitem__.return_value = 'mockingbird'
+        source.get.return_value = 'mockingbird'
+        helpers_mock.validate_remote_url.return_value = False
+        written_source = writers.write_source(conn, source)
+        helpers_mock.validate_remote_url.assert_called_with('mockingbird')
+
+        assert written_source == None
```
vitorbaptista: Same, just change to `assert written_source is None`

tests/processors/base/writers/test_source.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+class TestSourceWriter(object):
+    @mock.patch('processors.base.writers.source.helpers')
+    def test_skips_source_with_invalid_url(self, helpers_mock):
+        conn = _get_mock_conn()
+        conn['database']['sources'].find_one.return_value = None
+        source = mock.MagicMock(spec_set=dict)
+        source.__getitem__.return_value = 'mockingbird'
+        source.get.return_value = 'mockingbird'
```
vitorbaptista: Isn't the `source` just a regular `dict`? In that case, there's no need to use a mock here, you can simply pass a regular dict with the values you want.

tests/processors/base/writers/test_source.py
```diff
@@ -0,0 +1,46 @@
...
+
+        assert written_source == None
+
+
+    @mock.patch('processors.base.writers.source.helpers')
+    def test_writes_source_with_valid_url(self, helpers_mock):
+        conn = _get_mock_conn()
+        conn['database']['sources'].find_one.return_value = None
+        source = mock.MagicMock(spec_set=dict)
+        source.__getitem__.return_value = 'mockingbird'
```
vitorbaptista: Same as above, no need to use a mock to replace a dict.

tests/processors/base/writers/test_document.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+    @mock.patch('processors.base.writers.document.helpers')
+    def test_writes_document_with_valid_url(self, helpers_mock):
+        conn = _get_mock_conn()
+        conn['database']['documents'].find_one.return_value = None
+        document = mock.MagicMock(spec_set=dict)
+        document.__getitem__.return_value = 'mockingbird'
```
vitorbaptista: Replace mock with regular dict

tests/processors/base/writers/test_document.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+class TestDocumentWriter(object):
+    @mock.patch('processors.base.writers.document.helpers')
+    def test_skips_document_with_invalid_url(self, helpers_mock):
+        conn = _get_mock_conn()
+        conn['database']['documents'].find_one.return_value = None
+        document = mock.MagicMock(spec_set=dict)
+        document.__getitem__.return_value = 'mockingbird'
+        document.get.return_value = 'mockingbird'
```
vitorbaptista: Replace mock with regular dict.

processors/base/writers/source.py
```diff
@@ -48,6 +49,21 @@ def write_source(conn, source):
         'terms_and_conditions_url': source.get('terms_and_conditions_url'),
     })
 
+    # Validate object
+    url_fields = ['url', 'terms_and_conditions_url']
+    failed_url_validation = [field for field in url_fields
+                             if not helpers.validate_remote_url(object[field])]
+    if failed_url_validation:
+        for field in failed_url_validation:
```
vitorbaptista: Instead of printing multiple log messages, print only one passing the actual `failed_url_validation` array. It'll be converted to string anyway, and makes our logs less noisy.

processors/base/helpers/__init__.py
```diff
@@ -163,6 +164,19 @@ def validate_identifier(identifier):
             return int(numbers) != 0
 
 
+def validate_remote_url(url):
+    """"URLs without domain and scheme or with a file scheme are invalid"""
+    is_valid_url = False
+    try:
+        parsed_url = urlparse.urlparse(url)
+        if parsed_url.scheme and parsed_url.netloc:
+            is_valid_url = parsed_url.scheme != 'file'
```
vitorbaptista: To improve legibility, please add paranthesis here (i.e. `is_valid_url = (parsed_url.scheme != 'file')`)

vitorbaptista (changes requested):
All good, my review is mostly for minor Python-specific details (i.e. `foo is None` instead of `foo == None`) and to reduce the mocking.

Mocking is a very important tool, but its use should be kept to a minimum. It's very easy to end up mocking everything and ending up with tests that are too brittle (i.e. break unnecessarily all the time) or keep passing even if the code is broken.

Ideally, if our code was testable without many external dependencies, we wouldn't need to mock anything. It isn't yet (we don't have a test database, for example), but we'll get there.

vitorbaptista:
I understand the reasoning on mocking `helpers.validate_remote_url` on the writers' tests, so the test won't depend on the implementation of that method. However, it creates a drawback: this test now depends on that method existing and being used. If/when we refactor the helpers, we'll need to change not only the helpers themselves and who calls  them (all good and expected), but also all those tests.

If, instead of testing _how_ the code is doing it's job, we focus on the outputs, that won't change. So the tests could be simplified to something like:

``` python
def test_writes_document_with_valid_url(self):
    document = {
      url: 'http://example.org',
       // ...
    }
    conn = _get_mock_conn()
    conn['database']['documents'].find_one.return_value = None

    assert writers.write_document(conn, document) is not None

def test_skips_documents_with_invalid_url(self):
    document = {
      url: 'invalid-url',
       // ...
    }
    conn = _get_mock_conn()
    conn['database']['documents'].find_one.return_value = None

    assert writers.write_document(conn, document) is None
```

Notice we're not even testing that the `write_document` returned the `id` of the document, but only that it was created, which is the focus of these tests. This simplifies the test code quite a bit, and makes them easier to maintain (as they won't be breaking unnecessarily).

Also, you just wrote tests for sources and documents, but the code also changes records, publications, and file. Please add tests for the others.

vitorbaptista (approved):

vitorbaptista:
Great job! Happy to see we'll eliminate a source of data bugs :+1:

PR #70, opened 2016-10-06
[#451] Add processors to remove sources

vitorbaptista:
I needed to add two processors because this process takes multiples steps. One
of them deletes all sources' data except the trials. This is because a trial
can have multiple sources, including sources that we don't want to remove. We
can't simply delete them, otherwise we would loose their IDs, so the process
has to be:
1. Run remove_sources_data to delete all data from the sources except the
   trials;
2. Run all other sources' processors to update the remaining trials with
   the data from sources we won't remove;
3. Run trial_remover to remove trials without records;
4. Run sources_remover to finally remove the sources themselves.

Fixes opentrials/opentrials#451


victor-nitu:
The reverse operation for this would be to run the corresponding processor(s), right?

I haven't found anything suspicious, so I guess this is ready to be merged. I even learned a thing or two about the libs we're using while reading this :)

Let's merge and run it 👍

PR #73, opened 2016-10-21
[#507] Documents and Trials now have a many-to-many relationship

vitorbaptista:
opentrials/opentrials#507

Merge https://github.com/opentrials/processors/pull/74 before this, as it depends on that.


processors/base/writers/document.py
```diff
@@ -70,14 +58,69 @@ def write_document(conn, document):
...
+
+        db['documents'].upsert(obj, ['id'], ensure=False)
+
+        trial_id = document.get('trial_id')
+        if trial_id:
+            trials_documents = {
+                'trial_id': trial_id,
+                'document_id': obj['id'],
+            }
+            db['trials_documents'].insert_ignore(
```
georgiana-b, line 73: Just out of curiosity, why do we need insert_ignore here?

georgiana-b:
Glad to see an example with real db tests. Seems fine to me.

processors/base/writers/document.py
vitorbaptista: `insert_ignore` inserts a new row if one with the keys doesn't exist (here `['trial_id', 'document_id']`), or ignores if one exist (i.e. doesn't update or do anything). In contrast with the more common `upsert`, which updates an existing row if it exists, `insert_ignore` does nothing in that case.

This is because `trials_documents` only has columns `trial_id` and `document_id`, so updating it doesn't make sense (it would update it with the same values it has now, not changing anything)

PR #74, opened 2016-10-21
[#519] Fix bug where sources and documents required URL

vitorbaptista:
Their URLs are nullable.

Fixes opentrials/opentrials#519


processors/base/writers/document.py
```diff
@@ -60,7 +60,7 @@ def write_document(conn, document):
     })
 
     # Validate object
-    if not helpers.validate_remote_url(obj['url']):
+    if obj.get('url') is not None and not helpers.validate_remote_url(obj['url']):
```
georgiana-b: If the mystery of how documents get this far without a URL gets solved and we are going to make this check then we should remove [its counterparts from `extract_documents`](https://github.com/opentrials/processors/blob/7765570748641b24b8dc128bbee65c230765e5b5/processors/euctr/extractors.py#L132).

georgiana-b:
Seems ok to me.

PR #75, opened 2016-10-24
[#522] Rename url column for documents, files and sources

georgiana-b:
opentrials/opentrials#522


vitorbaptista:
_Update: @georgiana-b and I talked and this change isn't needed._

So, we're missing updating the tests for `processors.base.writers.source`. However, this change exposed a failure in our test suite: it should have broken it, but it didn't. The problem is that on https://github.com/opentrials/processors/blob/917a231e5b6e55c59418ad254102d78a63449cb2/tests/processors/base/writers/test_source.py#L12-L23 we're testing both invalid `url` and `terms_and_conditions_url` so, even though `url` isn't validated anymore, `terms_and_conditions_url` is, which makes the test pass (even though it shouldn't).

To fix this, please change `test_skips_source_with_invalid_url` to have only an invalid `source_url` (renaming the method name as well), and no `terms_and_conditions_url` at all (as it isn't required). Then add a new test `test_skips_source_with_invalid_terms_and_conditions_url`, this time without a `source_url`.

PR #76, opened 2016-10-25
Extract submitting documents do DocumentCloud from FDA_DAP processor

vitorbaptista:
The FDA_DAP processor was responsible for too many things (downloading PDFs, merging PDFs, uploading to S3, submitting to DocumentCloud). If we needed to, say, add more metadata to the documents in DocumentCloud, we'd have to change and run the FDA_DAP processor again, which does all those things (including downloading/merging PDFs), and so is much slower than needed.

This PR extracts the DocumentCloud-specific parts into a separate `send_fda_docs_to_documentcloud` processor, so it's more manageable and can be run quicker.

Fixes opentrials/opentrials#505

### DIFF
--- a/docker-cloud.yml
+++ b/docker-cloud.yml
@@ -346,6 +346,21 @@ sync-text-from-documentcloud:
   tags:
     - opentrials
 
+send-fda-docs-to-documentcloud:
+  image: okibot/processors
+  command: make start send_fda_docs_to_documentcloud
+  # restart: always
+  environment:
+    WAREHOUSE_URL:
+    DATABASE_URL:
+    EXPLORERDB_URL:
+    LOGGING_URL:
+    DOCUMENTCLOUD_USERNAME:
+    DOCUMENTCLOUD_PASSWORD:
+    DOCUMENTCLOUD_PROJECT:
+  tags:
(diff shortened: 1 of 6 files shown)


victor-nitu:
All looking really good, just one observation/question: are we going to settle on this way of creating new processors (which I like) or just build them however we want as long as they have the required signature? I mean wrapping them into a class and all the structure in this one.

Not sure what would be the short term benefit (if any) of further standardizing the way we build the new processors, but it might pay off in the long run. But that's likely another discussion, just wanted to raise it here as it was convenient.

Feel free to merge it whenever you want, if you don't have anything to add to it! 👍

vitorbaptista:
@nightsh Obviously, I do prefer this way of writing the processors, as it's easier to create helper functions with a class, instead of passing `conn` and `conf` to all of them. We might end up using it as the default pattern, let's see...

victor-nitu:
I am thinking of a "maintenance" or "housekeeping" day if we want to make them all uniform. Just attempted a refactoring on the record remover, and realized the helpers are built for some slightly incompatible signatures. (requiring `conn` and `"database"` separately, for example).

PR #77, opened 2016-10-25
[#522] Expect documents, sources and files to have "source_url"

vitorbaptista:
Not "url", as it has a different meaning in our API.

opentrials/opentrials#522

@georgiana-b Even though we talked about keeping it as `url` to avoid changing too much code, I ended up doing it after looking for a bug caused by this change. The bug is because we use raw queries in quite a few places, and some of them were expecting an `url` (i.e. we forgot to fix them in the previous PR). This fixes those as well.


georgiana-b:
Missed [this one](https://github.com/opentrials/processors/blob/2851dc3886816a70cfdd4ef656fe9e029b94a789/processors/contrib/processor.py#L79).
Of [this](https://github.com/opentrials/processors/blob/2851dc3886816a70cfdd4ef656fe9e029b94a789/processors/openaire/processor.py#L23) I'm not sure because I can't tell who it refers to but it's worth double checking.

vitorbaptista:
@georgiana-b Nice catch. Added the contrib one. The OpenAIRE one with the big SQL query isn't relevant (i.e. it's already using `source_url`)

PR #80, opened 2016-10-31
Improve performance of remove_unknown_documentcloud_docs

vitorbaptista:
We were getting the documents and then deleting it. We don't need to get the
document, we can delete it just using the ID we already have in our DB. This
reduces the number of API requests in half.


georgiana-b:
:+1: Good to go

PR #81, opened 2016-11-01
Bug/remove duplicate records from same url

victor-nitu:
[#532] Remove duplicate records with the same `source_url`

* extend `record_remover` processor
* Fixes opentrials/opentrials#532

Also refactor `trial_remover` into a class, coding style changes only.


victor-nitu:
@vitorbaptista I haven't been able to test it properly. I expect this to be a lot easier once we deal with opentrials/opentrials#526

processors/record_remover/processor.py
```diff
@@ -12,52 +12,97 @@
...
+        """
+
+        # Identify duplicates by `source_url` from the records DB
+        query = """
+        SELECT r1.id, r2.id, r1.source_url
+        FROM records AS r1
+        INNER JOIN records AS r2
+            ON r1.source_url = r2.source_url
+        AND r1.id != r2.id
+        AND r1.id > r2.id
```
vitorbaptista: Missing identation here. Also, as we're only using `r1.id`, you can remove `r2.id` and `r1.source_url` from the SELECT clause. They're only useful on the JOIN.

processors/record_remover/processor.py
```diff
@@ -12,52 +12,97 @@
...
+                    if self._conn['database']['records'].find_one(id=record['id']):
+                        self._conn['database']['records'].delete(id=record['id'])
+                        logger.info('Removed record: %s', record['identifiers'])
+
+            # Log info
+            if count and not count % 100:
+                logger.info('Processed %s records', count)
+
+
+    def _remove_records_from_same_source(self):
```
vitorbaptista: The name is confusing. It's not removing records from the same source, but removing records with the same source_url.

vitorbaptista:
@nightsh All good, just minor indentation and naming fixes :+1:

PR #82, opened 2016-11-02
Extract each individual page from DocumentCloud

vitorbaptista:
This depends on a migration on the related branch in the API, turning the `files.text` field into `files.page` with type `text[]`

opentrials/opentrials#500


tests/processors/base/writers/test_file.py
```diff
@@ -31,6 +31,22 @@ def test_writes_file_with_valid_url(self):
...
+            'source_url': 'http://example.org',
+            'documentcloud_id': '1000',
+            'pages': ['page 1', 'page 2'],
+        }
+        conn = _get_mock_conn()
+        conn['database']['files'].find_one.return_value = file_record.copy()
+
+        writers.write_file(conn, {'id': file_record['id']})
+
+        upsertted_file = conn['database']['files'].upsert.call_args[0][0]
```
victor-nitu: "upserted"

victor-nitu (approved):
All looks good, I also tested it locally. After we sort out opentrials/api#115 we can merge it right away.

PR #83, opened 2016-11-02
[#512] Allow URLs in EUCTR trial results

georgiana-b:
opentrials/opentrials#512

I'm not sure why the previous condition was `if record['trial_results'] == 'View results':` since EUCTR collector [extracts a URL for `trials_results`](https://github.com/opentrials/collectors/blob/master/collectors/euctr/parser.py#L58), but I wrote a condition that allows both.


vitorbaptista:
In the warehouse, there're 25,510 rows with `trial_results == 'View results'` and 1,474 with `trial_results == 'Removed from public view'`. Probably this is because we haven't re-run the collectors/processors on EUCTR after some changes. Could you confirm it?

The way I'd approach it is getting some EUCTR row with a non-url `trial_results` and add a test in https://github.com/opentrials/collectors/blob/master/tests/collectors/euctr/test_parser.py that validates that, for them, the `trial_results` will be null.

georgiana-b:
The collector was still returning "Removed from public view" when there was no results URL for `trial_results`. I fixed it and added a test in opentrials/collectors#51.

vitorbaptista:
Great addition, specially the tests! :clap: Happy to see that we're starting to improve our test suite.

There's just a quick improvement for the tests I'd like to before merging, specifically methods `test_extract_document_if_trial_results_url`, `test_trial_has_published_results_true_if_trial_results_url`, and `test_extracted_identifiers_for_trial`.

Ideally, each test is self-contained (i.e. it doesn't depend on anything outside of it). Sometimes it's too difficult, and we do things like `before`/`after` callbacks. However, for those tests it's not, and you've done it on the other methods. All of those tests assume that `self.STUB_RECORD` has some specific values. For example, `test_extracted_identifiers_for_trial` extracts data from the stub record and test which identifiers were extracted. This means that changing the values in `self.STUB_RECORD` will break this test. The fix is easy: just define explicitly all the values a test need. In that test, for example, you'd define `eudract_number` and `us_nct_clinicaltrials_gov_registry_number`. The others would have other definitions.

There's another small improvement here, but that's less important. You could change the `STUB_RECORD` to be a `@pytest.fixture`, simply doing something like:

``` python
@pytest.fixture
def stub_record():
    return {
        // ...
    }
```

This way you won't need to use `copy.deepcopy()`, as each test instance would get a new copy of the stub and can change it as necessary.

PR #84, opened 2016-11-03
[#546] Prepend to identifiers without making them invalid

georgiana-b:
opentrials/opentrials#546


victor-nitu (approved):
Since there's nobody assigned (yet?), I took the liberty to review it myself. I have nothing to add, it can be merged as far as I'm concerned.

tests/processors/base/helpers/test_helpers.py
```diff
@@ -116,3 +116,12 @@ def test_valid_url(self, url):
     ])
     def test_invalid_url(self, url):
         assert not helpers.validate_remote_url(url)
+
+def TestSafePrepend(object):
```
vitorbaptista: Tricky typo here: you're using `def TestSafePrepend()` instead of `class TestSafePrepend()`. This is why the last test `('EUCTR', None, 'EUCTR')` is passing (i.e. it's not running).

Also, if `helpers.safe_prepend()` received `None`, it should return `None` and not the `prepend_string` IMHO.

processors/base/helpers/__init__.py
```diff
@@ -250,3 +250,18 @@ def find_trial_by_identifiers(conn, identifiers, ignore_record_id=None):
...
+    """Prepend to string non-destructively
+    Ex: safe_prepend('EUCTR', 'EUCTR12345678') => 'EUCTR12345678'
+
+    Args:
+        prepend_string: string to prepend
+        string: string to prepend to
+    """
+
+    if not string.startswith(prepend_string):
+        string = '%s%s' % prepend_string, string
```
vitorbaptista: Please add parenthesis like `'%s%s' % (prepend_string, string)` to make it clearer.

vitorbaptista:
@georgiana-b good addition! And @nightsh that's definitely correct, we all should be reviewing each other's code.

I noticed a bug with this, though. See my comments :v:

PR #86, opened 2016-11-14
WIP: [#526] Use PostgreSQL as test database

georgiana-b:
opentrials/opentrials#526


tests/conftest.py
```diff
@@ -0,0 +1,82 @@
...
+    metadata.drop_all()
+
+
+def create_test_database(source_database_url, test_database_url):
+    engine = create_engine(source_database_url)
+    test_engine = create_engine(test_database_url)
+    metadata = MetaData()
+    metadata.reflect(engine)
+    metadata.create_all(test_engine)
+    return test_engine
```
vitorbaptista: I don't think this approach will work on test environments like Travis, as they don't have access to a "regular" database. In that case, we need to generate the schema dumps locally and commit them in.

georgiana-b: I think this is caused by the fact that it comes from a fork. I noticed env variables have been added to the `opentrials/processors` repository which I don't have in my fork. 

```
Setting environment variables from repository settings

$ export DOCKER_EMAIL=[secure]

$ export DOCKER_USER=[secure]

$ export DOCKER_PASS=[secure]

$ export WAREHOUSE_URL=[secure]

$ export DATABASE_URL=[secure]

```

Travis fails with the error:

```
  File "/home/travis/build/opentrials/processors/tests/conftest.py", line 11, in <module>

    import processors.base.config as config

  File "/home/travis/build/opentrials/processors/processors/base/config.py", line 16, in <module>

    WAREHOUSE_URL = os.environ['WAREHOUSE_URL']

  File "/home/travis/virtualenv/python2.7.10/lib/python2.7/UserDict.py", line 23, in __getitem__

    raise KeyError(key)

KeyError: u'WAREHOUSE_URL'
```

I'll make a PR from a branch here after I'm done with the functionality.

vitorbaptista: I haven't checked the current Travis build, just saying this because by loading the schema like this will require that, to run the tests, you need all OpenTrials' databases. This makes it more difficult not only for Travis, but also for people cloning our repositories and wanting to set it up. It would also make the tests slower, as it dumps the schema every time you run the tests.

Because of this, I think that loading the schema on demand (i.e. when the tests are run) creates more problems than they solve, so aren't worth it.

Please just create a script that, when run, will dump the schema and write them to some place (e.g. `tests/dbs`). The tests just need to clean the data in the DB, dumping and loading the schema from the other DBs should be separate tasks, not run on each test run.

PR #90, opened 2016-12-09
Add fda_linker processor to docker cloud config

victor-nitu:


victor-nitu:
Will merge this now, as we need it to run the FDA linker in docker cloud. Please take a look when you can, it has pretty much the same config as the other images.

PR #91, opened 2016-12-09
[#602] Temporary processor for removing duplicates from EUCTR warehouse

georgiana-b:
opentrials/opentrials#602

### DIFF
--- a/processors/euctr_duplicates_remover/processor.py
+++ b/processors/euctr_duplicates_remover/processor.py
@@ -1,0 +1,28 @@
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import logging
+logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


processors/euctr_duplicates_remover/processor.py
```diff
@@ -0,0 +1,28 @@
...
+import logging
+logger = logging.getLogger(__name__)
+
+
+# Module API
+
+def process(conf, conn):
+    """Remove duplicate records from EUCTR"""
+
+    query = "SELECT eudract_number_with_country from euctr where eudract_number_with_country ~ '^EUCTR*';"
```
victor-nitu, line 16: So to make things clear, initially we had `EUCTR` as part of the ID, then we dropped it but also kept the old records instead of removing them. This led to a duplication of some EUCTR records, which is why we have to remove the old ones from the database.

victor-nitu (approved):
Looking good, merge it and let's make it run 👍

PR #95, opened 2016-12-14
Feature/mark primary record

georgiana-b:
opentrials/opentrials#610

### DIFF
--- a/processors/base/processors/trial.py
+++ b/processors/base/processors/trial.py
@@ -40,8 +40,16 @@ def process_trials(conn, table, extractors):
             if trial_id is None:
                 continue
 
+            # Set current primary record to false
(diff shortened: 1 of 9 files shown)


victor-nitu (approved):
Looking good 👍 
(just wait for the [corresponding API PR](https://github.com/opentrials/api/pull/131) to be merged first)

victor-nitu:
About the changes in `tests/processors/cochrane_reviews/test_processor.py`, are they intentional? It's not related to the scope of this PR, but in case it's a hotfix for the scraper then it's ok.

georgiana-b:
It is a hotfix. That test was failing because the previous record was removed from Pubmed.

PR #96, opened 2016-12-15
[#594] Add last_verification_date to record

georgiana-b:
opentrials/opentrials#594

The migration is in opentrials/api#132 which should be reviewed & deployed first.


victor-nitu (approved):
Approved, assuming only `gsk` and `nct` have last verification date fields to extract.

PR #98, opened 2016-12-24
[WIP] [#250] Add location normalization using HAC + constraints algorithm

felipevieira:
We are using an Hierarchical Agglomerative Clusteing (HAC) algorithm (more in https://en.wikipedia.org/wiki/Hierarchical_clustering) with several relation constraints to cluster all location variations found on the database, this way we can use these pre-built clusters to suggest new location normalizations (if that's the case). The main advantage of using this clustering approach is that, besides clustering strings by their edition distances, custom constraints can help us to add corner case rules that can complete our normalizations. For the sake of information, this is the locations partition (and its clusters) found when using the clustering algorithm (as it is)
countries_cannotlink_canonical_mustlink_acronym
Even though this dendrogram might seem hard to understand, it is also easy to infer that different colors indicate different clusters of location variations (e.g. [Thaland, Thaoland, Thaialnd, Thailand, Thailland]). The black line crossing the plot is the distance threshold (0.15) we've used to tell the algorithm to stop clustering.

Currently the algorithm is only using as clustering metric the Jaro–Winkler Distance, and by using only this attribute our results got an accuracy of ~70% over all the locations stored in database. However, our team is also planning to increment this clustering model to also cover different string edition distances, as well as geographical (e.g. cluster "Washington" and "United States" together) and semantic (e.g. cluster "All Scandinavian countries" with "Sweden", "Finland" and so on...) elements. At the end of this incremental model evolution process we might ensure that our clusters are as closest as possible to cover all variations that might be processed, guaranteeing that all location will be normalized as their canonical forms

https://github.com/opentrials/opentrials/issues/250


pwalsh:
Doesn't seem to be running against the database / the table that holds location data.

felipevieira:
@pwalsh Would you please tell me how did you perform your tests so I can reproduce your issue?

pwalsh:
@felipevieira I did my test by reading the code - I can't see how this processors works on the warehouse database to normalise location data.

felipevieira:
@pwalsh In our tests the successful normalization happens on the location writter, where instead of considering the raw location name provided by the warehouse, the writter asks our HAC model for a normalizated form of that location (line 37 of the writter). This is how we thought the normalization would work.

Anyway, we are aware that this solution as it is does not normalize the full location database. Instead, it only ensures that new processed locations will be submitted to the normalization process. Our team is working to normalize the full database as we speak

PR #99, opened 2016-12-29
[#625] Remove trial relations before removing trial

georgiana-b:
opentrials/opentrials#625

### DIFF
--- a/processors/trial_remover/processor.py
+++ b/processors/trial_remover/processor.py
@@ -11,39 +11,129 @@ logger = logging.getLogger(__name__)
 # Module API
 
 def process(conf, conn):
-    """Remove trials without records.
+    """Removes trials alongs with their related entities (ex: documents)
+     if they are not related to any other trial or FDA application.
+
+    NOTE: Please run `remove_unknown_documentcloud_docs` processor after this processor.
     """
 
-    processor = _TrialRemover(conf, conn['database'])
-    processor.remove_trials_without_records()
+    remove_trials_without_records(conf, conn)
+
+
+def remove_trials_without_records(conf, conn):
+    """Remove trials without records"""
+
+    # Prepare
+    query = """
+        SELECT trials.id, trials.identifiers
+        FROM trials LEFT JOIN records ON records.trial_id = trials.id
+        WHERE records.trial_id IS NULL AND trials.source_id != 'pubmed'
+    """
+
+    # Execute
+    count = 0
+    for trial in conn['database'].query(query):
+        trial_id = trial['id'].hex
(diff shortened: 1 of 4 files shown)


processors/trial_remover/processor.py
```diff
@@ -11,39 +11,129 @@
...
+            logger.exception('Can\'t delete trial: %s and its relations due to error: %s',
+                             trial['identifiers'], repr(exception))
+        else:
+            conn['database'].commit()
+            logger.info('Deleted trial without records: %s and its relations',
+                        trial['identifiers'])
+            count += 1
+        if count and not count % 100:
+            logger.info('Removed %s trials without records', count)
+    logger.info('Removed %s trials without records', count)
```
victor-nitu, line 62: Should we also output the number of failed attempts (if any)? Just to know, at the end of it, if there's anything worth grepping through the log.

processors/trial_remover/processor.py
```diff
@@ -11,39 +11,129 @@
...
-                logger.exception('Can\'t delete trial: %s', trial['identifiers'])
-            else:
-                logger.info('Deleted trial without records: %s', trial['identifiers'])
-                count += 1
-            if count and not count % 100:
-                logger.info('Removed %s trials without records', count)
-        logger.info('Removed %s trials without records', count)
+            # If the related entity is not attached to other trials, delete it too
+            if len(entity_relations) == 0:
+                yield entity
```
victor-nitu, line 143: 👍

victor-nitu (approved):
Code really looks great, feel free to merge whenever you want.

As far as logging goes, I know we usually don't log the number of fails. It would probably be a good thing to do so – if you agree, please see my comment on `remove_trials_without_records` definition.

processors/trial_remover/processor.py
georgiana-b: @nightsh See line 58 and 59: If not all the trials (100%) are removed, it will say that only x% of them were.

processors/trial_remover/processor.py
```diff
@@ -58,6 +60,7 @@ def remove_trials_without_records(conf, conn):
         if count and not count % 100:
             logger.info('Removed %s trials without records', count)
     logger.info('Removed %s trials without records', count)
+    logger.info('Failed to remove %s trials without records', error_count)
```
victor-nitu: Better to have a message like:
> Removed 32145 trials without records (0 failed).

Or:
> Removed 32142 trials without records (3 failed).

WDYT, does it look better this way?

Or just print the 2nd message only if there are any fails, and with `logger.warn` instead.

PR #102, opened 2017-01-17
[WIP] [#250] Add support to a simple location normalization

felipevieira:
By combining ISO-3166 standards and the Levenshtein edition distance the processors now can (manually using a external csv file provided by Github) try to detect the canonical form of any non-normalized location stored on the warehouse records.

This normalization process as currently implemented was ran over all the location variations present on the OpenTrials database (293 different locations, according to [this file](https://github.com/opentrials/processors/files/369092/opentrials_locations.csv.zip) provided by @vitorbaptista in other discussion) and we were able to normalize **81.91%** of all occurrences. Most of the non-normalized cases can be described as corner-cases and (as stated by the issue description) should not be dealt (although they can be easily added to the auxiliary csv file to match specific locations, and that's the reason why we keep the csv file on the repository instead of accessing a remote file on-the-fly). We can provide a list of all normalizations performed by the algorithm, if desired.

Fixes https://github.com/opentrials/opentrials/issues/250


tests/processors/base/helpers/test_helpers.py
```diff
@@ -125,3 +125,83 @@ class TestSafePrepend(object):
...
+        ("UK","United Kingdom"),
+        ("UY","Uruguay"),
+        ("VAT","Holy See"),
+        ("YE","Yemen"),
+        ("KOR","Korea, Republic of"),
+        ("PE","Peru"),
+        ("NGA","Nigeria"),
+        ("ML","Mali"),
+
+        # Locations normalized by ISO-3166 capital standards
```
vitorbaptista: Don't alter city names to country names, as we don't want to lose information.

processors/base/helpers/__init__.py
```diff
@@ -4,20 +4,30 @@
...
 import re
 import json
 import uuid
 import string
 import logging
 import datetime
 import urlparse
+from iso3166 import countries
+from fuzzywuzzy import fuzz
+from regex import sub
```
vitorbaptista: Prefer using `import foo` and then `foo.bar()`, instead of `from foo import bar`. It improves the code readability (and "greppability"), as when seeing `regex.sub()` I know it's a method from `regex` (compare that with seeing `sub()` and not knowing where it came from)

processors/base/helpers/__init__.py
```diff
@@ -266,3 +279,43 @@ def safe_prepend(prepend_string, string):
...
+
+    clean_string = lambda u: sub(ur"\p{P}+", "", u).lower()
+    # Try to fetch canonical name directly from the iso-3166 country standard
+    try:
+        return countries.get(clean_string(location)).name
+    # If country isn't found on iso-3166 country list
+    except KeyError:
+        normalized = False
+        with open(os.path.join(os.path.dirname(__file__),
+                               'data/countries.csv'), 'r') as csv_file:
```
vitorbaptista: Move the file path to a constant at the beginning of the method

processors/base/helpers/__init__.py
```diff
@@ -266,3 +279,43 @@ def safe_prepend(prepend_string, string):
...
+            # Store information about the current match
+            current_match, current_distance = location, float("-inf")
+            for country in csv_file:
+                country_infos = unicode(country, encoding="utf-8").split(",")
```
vitorbaptista: Instead of parsing the CSV manually, use `csv.DictReader(csv_file)`. This way you don't need to keep track of the indexes of the columns you want.

processors/base/helpers/__init__.py
```diff
@@ -266,3 +279,43 @@ def safe_prepend(prepend_string, string):
...
+                lev_distances = [fuzz.ratio(clean_string(location),
+                                            country_info.lower()) for country_info in relevant_info]
+                # Update current match if any distance is above the threshold
+                # and the current max score
+                if max(lev_distances) >= EDIT_DISTANCE_THRESHOLD and \
+                    max(lev_distances) > current_distance:
+                    normalized = True
+                    current_match, current_distance = countries.get(
+                        country_infos[COUNTRY_ALPHA3_INDEX]).name,\
+                        max(lev_distances)
```
vitorbaptista: This line break makes this very hard to read. Instead of using this multiple assignment, simply place one per line.

Also, instead of keeping track of the `normalized` boolean, you could just keep track of the higher ratio and match and, outside of this loop, do a test like:

```python
if current_distance >= EDIT_DISTANCE_THRESHOLD:
    # Accept the match
else:
    # Reject it
```

This keeps the logic of finding the best match separated from accepting it or not.

processors/base/helpers/__init__.py
```diff
@@ -266,3 +279,43 @@ def safe_prepend(prepend_string, string):
...
+            # Store information about the current match
+            current_match, current_distance = location, float("-inf")
+            for country in csv_file:
+                country_infos = unicode(country, encoding="utf-8").split(",")
+                # Calculate levenshtein distance between the passed country
+                # and all country relevant informations
+                relevant_info = country_infos[0:5] + \
+                    [country_infos[COUNTRY_CAPITAL_INDEX]]
+                lev_distances = [fuzz.ratio(clean_string(location),
+                                            country_info.lower()) for country_info in relevant_info]
```
vitorbaptista: The `lev_distances` name is confusing, as it doesn't contain the actual levenshtein distance, but a kind of similarity ratio based on it. Also, instead of breaking the line in the middle of the method call, do something like:

```python
foo = [fuzz.ratio(bar, baz)
       for baz in relevant_info]
```

You'll need to be a bit creative to do it while keeping the line length, like saving the cleaned location name in another variable (which would save a few hundred calculations as a side benefit), but shouldn't be too hard.

processors/base/helpers/__init__.py
```diff
@@ -266,3 +279,43 @@ def safe_prepend(prepend_string, string):
...
+
+def get_canonical_location_name(location):
+    """Find the canonical location name according to the
+    passed entry
+
+    Args:
+        location (str): the location to be normalized
+    """
+
+    clean_string = lambda u: sub(ur"\p{P}+", "", u).lower()
```
vitorbaptista: If you're just removing punctuation, `u.translate(None, string.punctuation)` is better (and avoids adding a new dependency).

vitorbaptista (changes requested):
Please check my comments inline. Apart from them, the `fuzzywuzzy` library you're using has the functionality to extract the `n` best matches for a given string (https://github.com/seatgeek/fuzzywuzzy#process). As far as I can see, you could simplify the process to:

1. Load the data from the CSV
2. Use `fuzzywuzzy.process.extractOne()` to get the best match
3. Check if it's good enough and act accordingly

Possibly using just one or a few of the columns in the `countries.csv` file.

Talking about that file, it's also important to document where it came from and how to update it.

In general, avoid doing things like `a, b = c, d` unless you're extracting data from an array (i.e. `a, b = arr[0:2]`) or switching variable's values (i.e. `a, b = b, a`), as it increases the reading complexity a bit.

processors/base/helpers/__init__.py
felipevieira: Resolved

processors/base/helpers/__init__.py
felipevieira: Resolved. Your code couldn't be used since it's designed to string (I'm using unicode), but I've tried a similar approach without adding a new depedency

processors/base/helpers/__init__.py
felipevieira: Resolved

processors/base/helpers/__init__.py
felipevieira: Resolved

processors/base/helpers/__init__.py
felipevieira: Resolved

processors/base/helpers/__init__.py
felipevieira: Resolved

tests/processors/base/helpers/test_helpers.py
felipevieira: Resolved

felipevieira (commented):
Thank you for your feedback @vitorbaptista. I've tried to fix all points listed on this review. Please let me know if there's some pending changes

processors/base/helpers/__init__.py
```diff
@@ -266,3 +277,56 @@ def safe_prepend(prepend_string, string):
...
+        location (str): the location to be normalized
+    """
+
+    # This auxiliar file is provided by the dataset repository of github.com and
+    # can be checked on its latest version on the project website
+    # See: https://github.com/datasets/country-codes/blob/master/data/country-codes.csv
+    #
+    # To create e new entry (or update an existing one), add a new line (or find the
+    # existing one) and fill in the informations according to the header (line 1) order,
+    # separating fields using comma
```
vitorbaptista: No need to be too verbose here. Simply `Extracted from https://github.com/datasets/country-codes/blob/master/data/country-codes.csv` is enough.

processors/base/helpers/__init__.py
```diff
@@ -266,3 +277,56 @@ def safe_prepend(prepend_string, string):
...
+            current_score = float("-inf")
+            for country in reader:
+                # Extract relevant comparative info from the current country
+                relevant_info = [unicode(country[field], encoding="utf-8")
+                                 for field in reader.fieldnames[0:5]]
+                # Do some string clean up and extract the most similar field (according
+                # to Levenshtein scoring) over all the location relevant informations
+                cleaned_location = clean_string(location)
+                location_info = [location_info.lower() for location_info in relevant_info]
+                choice = process.extractOne(cleaned_location, location_info, scorer=DISTANCE_SCORER)
```
vitorbaptista: This is a good place to use multiple variable assignment, as `match, score = process.extractOne()`. This way, you avoid the `SCORE_INDEX`.

processors/base/helpers/__init__.py
```diff
@@ -266,3 +277,56 @@ def safe_prepend(prepend_string, string):
...
+    # can be checked on its latest version on the project website
+    # See: https://github.com/datasets/country-codes/blob/master/data/country-codes.csv
+    #
+    # To create e new entry (or update an existing one), add a new line (or find the
+    # existing one) and fill in the informations according to the header (line 1) order,
+    # separating fields using comma
+    CSV_PATH = 'data/countries.csv'
+    ALPHA_3_HEADER = 'ISO3166-1-Alpha-3'
+    DISTANCE_SCORER = fuzz.token_sort_ratio
+    SCORE_INDEX = 1
```
vitorbaptista: There's no need to add the `ALPHA_3_HEADER`, as it's used only once and the column name is as expressive as the constant name.

processors/base/helpers/__init__.py
```diff
@@ -266,3 +277,56 @@ def safe_prepend(prepend_string, string):
...
+                choice = process.extractOne(cleaned_location, location_info, scorer=DISTANCE_SCORER)
+
+                # Update current match if the distance score is above the current max score
+                if choice[SCORE_INDEX] > current_score:
+                    current_match = iso3166.countries.get(country[ALPHA_3_HEADER]).name
+                    current_score = choice[SCORE_INDEX]
+
+            if current_score < EDIT_DISTANCE_THRESHOLD:
+                logger.debug('Location "%s" not normalized', location)
+                return location
```
vitorbaptista, line 327: Log also when the normalization is successful

tests/processors/base/helpers/test_helpers.py
```diff
@@ -125,3 +125,55 @@ class TestSafePrepend(object):
...
+        ("Virgin Islands (U.S.)","Virgin Islands, U.S."),
+
+        # Locations not normalized (corner cases)
+        ("Global trial(North America)","Global trial(North America)"),
+        ("Asia(except Japan)","Asia(except Japan)"),
+        ("Europe","Europe"),
+        ("Jiangsu","Jiangsu"),
+        ("Tianjin","Tianjin"),
+        ("Multinational","Multinational"),
+        ("Outside","Outside")])
```
vitorbaptista: We don't need so many examples of the same categories. Just keep a test of each and add some corner cases (e.g. what happens if it receives `None`?)

processors/base/helpers/__init__.py
```diff
@@ -266,3 +277,56 @@ def safe_prepend(prepend_string, string):
...
+            current_match = location
+            current_score = float("-inf")
+            for country in reader:
+                # Extract relevant comparative info from the current country
+                relevant_info = [unicode(country[field], encoding="utf-8")
+                                 for field in reader.fieldnames[0:5]]
+                # Do some string clean up and extract the most similar field (according
+                # to Levenshtein scoring) over all the location relevant informations
+                cleaned_location = clean_string(location)
+                location_info = [location_info.lower() for location_info in relevant_info]
```
vitorbaptista: I'd imagine it would be better to run the location data through the same `clean_string()` method we use for the location?

processors/base/helpers/__init__.py
```diff
@@ -266,3 +277,56 @@ def safe_prepend(prepend_string, string):
...
+    """
+
+    # This auxiliar file is provided by the dataset repository of github.com and
+    # can be checked on its latest version on the project website
+    # See: https://github.com/datasets/country-codes/blob/master/data/country-codes.csv
+    #
+    # To create e new entry (or update an existing one), add a new line (or find the
+    # existing one) and fill in the informations according to the header (line 1) order,
+    # separating fields using comma
+    CSV_PATH = 'data/countries.csv'
```
vitorbaptista: Set the constant to the entire path (i.e. `os.path.join(...)`)

processors/base/helpers/__init__.py
```diff
@@ -4,20 +4,28 @@
...
 import json
 import uuid
 import string
 import logging
 import datetime
 import urlparse
+import csv
+import iso3166
+
+from fuzzywuzzy import process, fuzz
```
vitorbaptista: Keep this import in the above section, just after the `iso3166`, and use `import fuzzywuzzy` instead of `from foo import bar`

vitorbaptista (changes requested):
Looking good, just needing some minor changes to be merged.

In general, avoid comments that just repeat what the code is doing, and use single quotes for strings (except when to avoid escaping).

processors/base/helpers/__init__.py
felipevieira: I've tried to do that after reading your first review, but for some reason I can't figure out (visibility?) I'm unable to do such thing as `import fuzzywuzzy` and then use `fuzzywuzzy.fuzz` or `fuzzywuzzy.process` (am I doing any nonsense?). Besides, the [fuzzywuzzy docs](https://github.com/seatgeek/fuzzywuzzy) actually uses this `from fuzzywuzzy import foo` format to import the sub-packages on its examples. That's why I'm keeping this import format so far. Any thoughts on how to keep the format you desire?

processors/base/helpers/__init__.py
felipevieira: Resolved

processors/base/helpers/__init__.py
felipevieira: Resolved

processors/base/helpers/__init__.py
felipevieira: Resolved

processors/base/helpers/__init__.py
felipevieira: Resolved

processors/base/helpers/__init__.py
felipevieira: Resolved. The reason why I hadn't done that yet is because since I'm not using the `match` variable (I'm fetching the current match directly from the iso3166 lib using the line information) I thought that would be a bad practice

processors/base/helpers/__init__.py
felipevieira: Resolved. I avoided logging successful normalizations in respect to @roll comment on the old PR, where he stated that logging all normalizations would generate a lot of noise on the logs. Instead, logging only the non-normalized cases would be more useful for audit purposes

tests/processors/base/helpers/test_helpers.py
felipevieira: Resolved

processors/base/helpers/__init__.py
vitorbaptista: In that case, you can do `_, score = process.extractOne()`. Naming a variable as `_` is a relatively well-known practice for saying "I don't care about this variable, but I need to add it"

processors/base/helpers/__init__.py
vitorbaptista: They don't automatically import the `fuzz` or `process` methods, so you'd need to do:

```python
import fuzzywuzzy.fuzz
import fuzzywuzzy.process
```

Also, remember of moving it to the import block close to `import csv`, `import iso3166`, etc.

processors/base/helpers/__init__.py
felipevieira: Didn't know about that. Thank you. Resolved.

processors/base/helpers/__init__.py
felipevieira: Resolved

felipevieira (commented):
All issues were resolved

processors/base/helpers/__init__.py
```diff
@@ -266,3 +279,47 @@ def safe_prepend(prepend_string, string):
...
+    if location is None:
+        return None
+
+    # Extracted from: https://github.com/datasets/country-codes/blob/master/data/country-codes.csv
+    CSV_PATH = os.path.join(os.path.dirname(__file__), 'data/countries.csv')
+    DISTANCE_SCORER = fuzzywuzzy.fuzz.token_sort_ratio
+
+    clean_string = lambda u: re.sub(u"[^\w\d'\s]+", '', u).lower()
+    try:
+        logger.debug('Location "%s" normalized using ISO-3166 standard', location)
```
vitorbaptista: You don't know if it was normalized here. The following line can still throw an exception. It's also good to log the normalized value, so you can change this to:

```python
normalized_location = iso3166.countries.get(cleaned_location).name
logger.debug('Location "%s" normalized as "%s"', cleaned_location, normalized_location)
return normalized_location
```

This also mean moving the `cleaned_location = clean_string(location)` from line 312 to before here, possibly before the `try` block.

processors/base/helpers/__init__.py
```diff
@@ -266,3 +279,47 @@ def safe_prepend(prepend_string, string):
...
+
+                if score > current_score:
+                    current_match = iso3166.countries.get(country['ISO3166-1-Alpha-3']).name
+                    current_score = score
+
+            if current_score < EDIT_DISTANCE_THRESHOLD:
+                logger.debug('Location "%s" not normalized', location)
+                return location
+
+            logger.debug('Location "%s" normalized using Levenshtein distance', location)
```
vitorbaptista: Log the normalized value as well, as `logger.debug('Location "%s" normalized as "%s"', location, current_match)`

vitorbaptista (changes requested):

processors/base/helpers/__init__.py
felipevieira: Sorry about that. Resolved

processors/base/helpers/__init__.py
felipevieira: Resolved

felipevieira (commented):
All issues were resolved

vitorbaptista (approved):

vitorbaptista:
@felipevieira All good! Could you just rebase this branch with master, so the build can run on Travis? If it's green, I'll merge it.

Thanks for your work :+1:

vitorbaptista:
@felipevieira Could you rebase again with master? I just merged some code at https://github.com/opentrials/processors/pull/108 that should make this pass on Travis.

felipevieira:
@vitorbaptista All green (y)

felipevieira:
@vitorbaptista All green :+1:

vitorbaptista:
@felipevieira Merged! Thanks for your contributions :beers:

PR #103, opened 2017-01-17
[WIP] [#353] Adds a simple approach to normalize study_phase values

arthurSena:
We opted to implement a solution based on a CSV file representing a two-way table(variation and its normalization). This file is accessed in order to match a phase variation
with its normalized forms. 

Since a Trial may be in one or more Phases simutaneously and, as it was suggested,we decided that the return of the method should be an array of phases. 
However, 'knex' does not support said type so we decided to alter the type of the 'study_phase' column to 'jsonb'. 

With this solution, whenever a new variation arrives the CSV file needs to be updated.
Furthermore, a new schema file was generated, updating tests/dbs.

With these modifications, changes were needed on the Api repository. In which the migration and seeds were updated, generating another [PR](https://github.com/opentrials/api/pull/136).


tests/processors/base/normalizers/test_phase_normalizer.py
```diff
@@ -0,0 +1,93 @@
...
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+from processors.base.normalizers import get_normalized_phase
+import pytest
+
+#Tests
+class TestValidateRemoteURL(object):
+    @pytest.mark.parametrize('test_input,expected', [
```
vitorbaptista: No need to test every possibility here. Try just a couple cases and the "extremes" (e.g. `None`, a phase we don't have the mapping for, etc.)

tests/processors/base/normalizers/test_phase_normalizer.py
```diff
@@ -0,0 +1,93 @@
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+from processors.base.normalizers import get_normalized_phase
+import pytest
```
vitorbaptista, line 7: The usual importing order we follow is:

1. Standard libraries
2. Third-party libraries
3. Our modules

processors/base/normalizers/phase_normalizer.py
```diff
@@ -0,0 +1,58 @@
...
+
+        :return:
+            phase_suggestions (list): normalized phase suggestions
+    """
+    if not phase:
+        logger.debug('Unsuccessfully phase normalization \'None\'')
+        return phase
+    phase_variation_map = build_phase_variation_map\
+                        (os.path.join(os.path.dirname(__file__),
+                                      'phases_variations.csv'))
```
vitorbaptista: Move the path to a constant, so we can avoid breaking the line here.

processors/base/normalizers/phase_normalizer.py
```diff
@@ -0,0 +1,58 @@
...
+        Else, returns the unormalized phase.
+
+        :param:
+            phase (str): unormalized phase
+
+        :return:
+            phase_suggestions (list): normalized phase suggestions
+    """
+    if not phase:
+        logger.debug('Unsuccessfully phase normalization \'None\'')
```
vitorbaptista: Remember that `phase` can be any falsy value, like `[]`, `()`, `0`, not only `None`, so it's better to log the `phase` value itself.

vitorbaptista (changes requested):
I think this is the way to go, there're just a couple changes and simplifications I'd like to do before merging.

The main simplification is, instead of saving the variations as a CSV and converting it to a dict, save it as a JSON like:

```json
{
    "Phase 1/2": [
         "Phase 1",
         "Phase 2"
    ],
    ...
}
```

This makes the code simpler (you can remove the `build_phase_variation_map()` method) and the variation file becomes more readable. Then, in the `get_normalized_phase()`, you can simply do:

```python
if phase in normalized_phases:
    result = normalized_phases[phase]
else:
    logger.debug('Unable to normalize phase', phase)
    if not phase:
        result = phase
    else:
        result = [phase]
return result
```

As we're only normalizing when we find an exact match, there's no need to log when we're successful.

vitorbaptista:
About using `jsonb` instead of `text[]`, have you tried using `table.specificType('study_phase', 'text[]')`? If so, did you have any problems with it?

arthurSena:
Yes, we did and it works fine. 
Acctually, we have already made the changes and we're going to commit them soon.

processors/base/normalizers/phases_variations.json
```diff
@@ -0,0 +1,245 @@
...
+  ],
+  "Phase4": [
+    "Phase 4"
+  ],
+  "n\\a": [
+    "Not applicable"
+  ],
+  "Phase 2": [
+    "Phase 2",
+    "Phase 2"
```
vitorbaptista, line 61: Are these repeating values a mistake? There are a few cases of this (e.g. on `Phase 3` and `Phase 1/Phase 2`)

processors/base/normalizers/phase_normalizer.py
```diff
@@ -0,0 +1,34 @@
...
+
+    with open(PHASE_VARIATION_PATH) as phase_variation_file:
+        phase_variation_map = json.load(phase_variation_file)
+
+    if phase in phase_variation_map.keys():
+        phase_suggestions = phase_variation_map[phase]
+    else:
+        logger.debug('Unable to normalize phase \'%s\'', phase)
+        if phase is None:
+            phase_suggestions = phase
```
vitorbaptista: Change this to:

```python
if phase:
    phase_suggestions = [phase]
```

And initialize `phase_suggestions = None` before line 26, so if `phase` is falsy (i.e. it's an empty string, `None`, an empty array, etc.) we would return None.

tests/processors/base/normalizers/test_phase_normalizer.py
```diff
@@ -0,0 +1,39 @@
...
+            ('IIIB', ['Phase 3B']),
+            ('0(exploratory trials))', ['Phase 0']),
+            ('I (Phase I study)', ['Phase 1']),
+            ('Post-market', ['Phase 4']),
+            ('Pilot study', ['Other']),
+            ('Diagnostic New Technique Clincal Study', ['Other']),
+            (None, None),
+            ('', ['']),
+            ('()', ['()']),
+            ('[]', ['[]']),
```
vitorbaptista, line 33: Not sure about values like `'()'` and `'[]'` (feel like a typo), but an empty string should return `None` as well. See https://github.com/opentrials/processors/pull/103/files#r97609345 for a suggestion on how to do so.

vitorbaptista (approved):

vitorbaptista:
@arthurSena thanks for the PR and congratulations on your first code merged :beers: 

Looking forward for the next ones :smile:

PR #104, opened 2017-01-23
[#467] Manage updated_at and created_at at db level

georgiana-b:
opentrials/opentrials#467

To be merged after opentrials/api#137

### DIFF
--- a/processors/base/writers/condition.py
+++ b/processors/base/writers/condition.py
@@ -6,7 +6,6 @@ from __future__ import unicode_literals
 
 import uuid
 import logging
-import datetime
 from .. import helpers
 logger = logging.getLogger(__name__)
 
@@ -30,7 +29,6 @@ def write_condition(conn, condition, source_id, trial_id=None):
 
     """
     create = False
-    timestamp = datetime.datetime.utcnow()
 
     # Get name
     name = helpers.clean_string(condition['name'])
@@ -45,7 +43,6 @@ def write_condition(conn, condition, source_id, trial_id=None):
     if not object:
         object = {}
         object['id'] = uuid.uuid1().hex
-        object['created_at'] = timestamp
         object['slug'] = slug
         create = True
 
@@ -54,7 +51,6 @@ def write_condition(conn, condition, source_id, trial_id=None):
 
         # Update object
         object.update({
-            'updated_at': timestamp,
             'source_id': source_id,
             # ---
             'name': name,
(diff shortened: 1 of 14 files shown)


tests/processors/merge_trials_identifiers/test_processor.py
```diff
@@ -31,7 +29,6 @@ def test_updates_trial_with_records_identifiers(self, conn, trial, record):
         updated_trial = conn['database']['trials'].find_one(id=trial)
 
         assert updated_trial['identifiers'] == record_attrs['identifiers']
-        assert updated_trial['updated_at'].replace(tzinfo=None) > trial_last_updated
```
georgiana-b: This was removed because the triggers that update `updated_at` were not cloned into the test database. If we want them in the test db we have to write and execute them again in here if we keep the SQLAlchemy-based schema dump or make the dump with `pg_dump`. Considering that we rarely test updates and when we do we can manually update the `updated_at` field, I figured it might not be worth the effort.
@vitorbaptista  What do you think?

vitorbaptista:
@georgiana-b Are you working on a trigger to update the `updated_at` column as well? If not, I think we can keep it as is, updating as we go, just because it's a bit difficult to keep track of database triggers in our codebase.

georgiana-b:
@vitorbaptista Yes, besides defaults at `INSERT` for both `created_at` and `updated_at`, I also made triggers to automatically update `updated_at` at `UPDATE`.
My question above is not whether or not we should keep the triggers but whether we should go through the trouble of cloning them into the test databases in `processors`.

tests/processors/merge_trials_identifiers/test_processor.py
vitorbaptista: Just to make sure I understood, the issue is that the code you've wrote to dump the database using SQLAlchemy doesn't copy the triggers, right?

It's tricky. In principle, the test database must be the same as the production database, otherwise we risk hitting hard (or impossible) to reproduce bugs. I went ahead and changed it to use regular postgres tools on https://github.com/opentrials/processors/pull/106. This simplified the code and surfaced other issues with SQLAlchemy not copying constraints. Please review that and then we can rebase it into this master before merging it in, so this test will continue working.

vitorbaptista: Just merged #106.

vitorbaptista:
@georgiana-b Please rebase with master and dump the new DB. Hopefully with the changes on https://github.com/opentrials/processors/pull/106 the test database will have everything the main database has.

victor-nitu:
@georgiana-b can you have a quick look at this please? I'd like a 2nd opinion, then hopefully we can move on and get all the trigger PRs merged.

victor-nitu (approved):

PR #105, opened 2017-01-23
[#645] Convert NCT's eligibility['gender'] == 'all' to 'both'

vitorbaptista:
ClinicalTrials.gov changed from using the term "Both" to using "All", which
makes us unable to import new data in our database, as we only allow genders to
be "male", "female" or "both".

Fixes opentrials/opentrials#645


vitorbaptista:
Merging this as we need it to update our data from NCT. Please review later.

PR #106, opened 2017-01-24
Dump DB schemas using psql

vitorbaptista:
Instead of using SQLAlchemy, use `pg_dump` and `psql` commands. This makes the code simpler and make sure we're don't miss things like indexes and constraints.

When merging this, squash it into a single commit.


georgiana-b (approved):
Good to go! Thanks for the help!

PR #107, opened 2017-01-25
[#653] Set Pfizer trial status as unknown when they're unknown

vitorbaptista:
**[#653] Set Pfizer trial status as unknown when they're unknown**
opentrials/opentrials#653


victor-nitu (approved):
Do it 😉

PR #108, opened 2017-01-25
[#658] [#421] Setup tox and require only test databases to run tests

vitorbaptista:
opentrials/opentrials#658
opentrials/opentrials#421


Makefile
```diff
@@ -6,9 +6,6 @@ all: list
 build:
 	docker build -t okibot/processors .
 
-install:
```
victor-nitu, line 9: any particular reason for removing this?

processors/base/config.py
```diff
@@ -10,14 +10,21 @@
...
-EXPLORERDB_URL = os.environ['EXPLORERDB_URL']
-TEST_WAREHOUSE_URL = os.environ.get('TEST_WAREHOUSE_URL')
-TEST_DATABASE_URL = os.environ.get('TEST_DATABASE_URL')
+if ENV == 'testing':
+    WAREHOUSE_URL = os.environ['TEST_WAREHOUSE_URL']
+    DATABASE_URL = os.environ['TEST_DATABASE_URL']
+else:
+    WAREHOUSE_URL = os.environ['WAREHOUSE_URL']
+    DATABASE_URL = os.environ['DATABASE_URL']
+    EXPLORERDB_URL = os.environ['EXPLORERDB_URL']
```
victor-nitu, line 27: I hope we're giving it the read only user, right?

victor-nitu (approved):

Makefile
vitorbaptista: In general, unless the commands aren't standard, I prefer to use the default Python tools. For instance, I prefer running `tox` instead of `make test`.

That's a personal preference, though, so you could argue both ways.

processors/base/config.py
vitorbaptista: As the tests are always run through `tox`, and `tox` itself sets the `PYTHON_ENV = 'testing'`, I think we're safe here. But yeah, if someone runs the tests outside of `tox` for some reason, she can end up clearing the local database :sweat:

Makefile
victor-nitu: I agree with that, just asking for the sake of clarifying the reason behind ;)

processors/base/config.py
victor-nitu: Can we do something (not only here, but in all code base) to abort tests if user has, for example, `DROP` or `TRUNCATE` privileges? Would be neat :)

vitorbaptista: I'm not sure if I follow. Why? During the tests we do clear the test database, to make sure they run in a clean state.

victor-nitu: Ah, never mind, I might have the "thinking out loud" syndrome...

Was thinking about a way to prevent accidents on explorer db – in this case – but trying to much to generalize it. Let's just forget about it 😅

PR #109, opened 2017-01-26
[#660] Setup Sentry

vitorbaptista:
opentrials/opentrials#660

PR #111, opened 2017-01-30
[#650] Replace document type with document categories

georgiana-b:
opentrials/opentrials#650

### DIFF
--- a/processors/euctr/extractors.py
+++ b/processors/euctr/extractors.py
@@ -132,8 +132,15 @@ def extract_documents(record):
     if results_url:
         document = {
             'name': 'Results',
-            'type': 'results',
             'source_url': results_url,
         }
         documents.append(document)
     return documents
+
+
+def extract_document_category(record):
+    return {
+        'id': 24,
+        'name': 'European Public Assessment Report (EPAR) document section',
+        'group': 'Results',
+    }
--- a/processors/base/processors/trial.py
+++ b/processors/base/processors/trial.py
@@ -54,10 +54,15 @@ def process_trials(conn, table, extractors):
             # Extract and write documents
             extract_documents = extractors.get('extract_documents')
             if extract_documents:
+
+                # Extract and write document category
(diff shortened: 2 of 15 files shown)


georgiana-b:
The Travis build is terminated because `make test` stalled, just as in #109, but I have no idea why. Any hypothesis would be appreciated.

processors/euctr/extractors.py
```diff
@@ -132,8 +132,15 @@ def extract_documents(record):
...
-            'type': 'results',
             'source_url': results_url,
         }
         documents.append(document)
     return documents
+
+
+def extract_document_category(record):
+    return {
+        'id': 24,
```
victor-nitu, line 143: Do we really need IDs for it?

victor-nitu (commented):
Looking at this family of related PRs on categories, I was thinking maybe we can upgrade to PostgreSQL 9.5 and use [the new UPSERTs](https://wiki.postgresql.org/wiki/What's_new_in_PostgreSQL_9.5#INSERT_..._ON_CONFLICT_DO_NOTHING.2FUPDATE_.28.22UPSERT.22.29). Pinging @vitorbaptista, does this affect us negatively in any imaginable way?

vitorbaptista:
About the upserts, we already have that functionality thanks to the `dataset` package. Upgrading postgres, as far as I can see, would only make it easier for `dataset` (actually, `sqlalchemy`) to implement this under the hood.

victor-nitu:
@georgiana-b try rebasing master onto this branch and force push, see if it fixes your travis builds.

processors/euctr/extractors.py
georgiana-b: `document_categories` will have to be kept in sink with Explorer's `data_categories` so we need a common identifier between them. If I choose `name` or [`name`, `group`] as `document_categories` unique identifier, when the `name` will change in `data_categories`, the corresponding `document_category` will not be updated but another one would be be created, leading to discrepancies. So I chose `data_categories` ID.

PR #115, opened 2017-02-09
[WIP] [#661] Deduplicate trials via title

arthurSena:
We opted to use the `public_title` approach in order to deduplicate trials as it doesn't have as much invalid titles as the `scientific_title`.
Also, we refactored the `helpers.find_trial_by_identifiers()` method to make it more readable and changed its signature so there is no need to call it twice in the `processors/base/writers/trial.py`. 
Please, let me know what you think about these changes.


processors/base/helpers/__init__.py
```diff
@@ -241,33 +241,58 @@ def iter_rows(conn, dataset, table, orderby, bufsize=100, **filter):
...
         if trial:
             break
+
+    if trial:
+        logger.debug('Trial-id %s was matched via identifiers with register-id %s',
+                     trial['id'], record_id)
+    elif source_id and trial_public_title:
+        trial_temp = conn['database']['trials'].find_one(
+                            public_title=trial_public_title)
+        if trial and trial.get('source-id') != source_id:
```
vitorbaptista: Is `source-id` a typo?

processors/base/helpers/__init__.py
```diff
@@ -241,33 +241,58 @@ def iter_rows(conn, dataset, table, orderby, bufsize=100, **filter):
             yield row
 
 
-def find_trial_by_identifiers(conn, identifiers, ignore_record_id=None):
-    """Find first trial matched by one of passed identifiers.
-
+def find_trial_by_identifiers(conn, identifiers, trial_public_title, source_id, record_id):
+    """Find if the trial already exists. Firstly, via identifiers. Secondly, via public title
```
vitorbaptista: The method name, `find_trial_by_identifiers`, doesn't make sense anymore, as we're now also using `public_title`. In general, I feel that this method should be left mostly unmodified, and a new method created that uses it. Something like `find_trial(conn, trial)` that would encapsulate the logic of finding it via identifiers, `public_title` or whatever else we use in the future. This would also mean that the only place you'll modify, apart from this file, is `processors/base/writers/trial.py`.

vitorbaptista (changes requested):

processors/base/helpers/__init__.py
arthurSena: Yes, you're right. I'm gonna make this changes.

arthurSena:
@vitorbaptista We are also commiting the audit log code in this PR which references the [#676](https://github.com/opentrials/opentrials/issues/676) issue. We thought that'd make sense as the another code depends on this one. Also, we made a [PR](https://github.com/opentrials/api/pull/147) on the api project that updates the migration to create the audit log for trial deduplication.

processors/base/helpers/__init__.py
```diff
@@ -271,6 +270,54 @@ def find_trial_by_identifiers(conn, identifiers, ignore_record_id=None):
...
+        dict: trial
+    """
+
+    def write_log(trial_id, record_id, method):
+        log = conn['database']['trial_deduplication_logs'].find_one(
+            trial_id=trial_id, record_id=record_id, method=method)
+
+        if not log:
+            repo = git.Repo(search_parent_directories=True)
+            commit_sha = repo.head.object.hexsha
```
vitorbaptista, line 290: Have you tried this on the processor Docker container? I haven't checked, but I'd imagine it'll have only the files themselves, and not the git repository.

processors/base/helpers/__init__.py
```diff
@@ -271,6 +270,54 @@ def find_trial_by_identifiers(conn, identifiers, ignore_record_id=None):
...
+        trial_found = find_trial_by_identifiers(conn, trial['identifiers'])
+
+    if trial_found:
+        logger.debug('Trial-id %s was matched via identifiers with register-id %s',
+                     trial_found['id'], record_id)
+        write_log(trial_found['id'], record_id, 'identifier')
+    elif source_id and trial.get('trial_public_title'):
+        trial_temp = conn['database']['trials'].find_one(
+                            public_title=trial.get('trial_public_title'))
+        if trial_temp and trial_temp.get('source_id') != source_id:
```
vitorbaptista: What if there're 2 trials with the same `public_title` but different sources? Using `find_one()`, you'd get only the first one, and ignore it if it has the same `source_id`, but maybe there is another one with a different `source_id` that we didn't get, because we just check the first one.

I think the best solution here is to do a raw query adding the `source_id`, like:

```python
query = '''
    SELECT * FROM trials
    WHERE public_title = :public_title
    AND source_id != :source_id
'''
conn['database']['trials'].query(
    query,
    public_title=trial.get('trial_public_title'),
    source_id=source_id
)
```

You could limit the results to 1 element by changing the query, as long as we have a way of handling trials currently in our database that should be removed thanks to the new deduplication process.

arthurSena: If I got it right, you mentioned that we should remove the trials that are currently duplicated in the database. In that case, is there any kind of source priority we should follow in order to do that? Or we could just keep the oldest trial in there and remove the other ones.

vitorbaptista: Consider the database as:

| id | public_title | source_id |
| --- | --- | --- |
| A | "A trial" | euctr |
| B | "A trial" | nct |

And a record with `{ public_title: "A trial", source_id: "euctr" }`. If we keep the code as is with the `.find_one()`, it would return the first trial (id: `A`), which would then be ignored as it has the same `source_id` as the record (i.e. `euctr`). The problem is that you're filtering the trials in two separate steps, one querying just for `public_title`, and with the results filter by `source_id`. One solution is to do both filters at once with the SQL query I suggested on https://github.com/opentrials/processors/pull/115#discussion_r103476928.

This is one issue. For the other issue, consider the database as:

| id | public_title | source_id |
| --- | --- | --- |
| A | "A trial" | euctr |
| B | "A trial" | nct |
| C | "A trial" | pfizer |

And the same record `{ public_title: "A trial", source_id: "euctr" }`. If we do a query for `public_title = 'A trial' AND source_id != 'euctr'`, we would get 2 results with sources `nct` and `pfizer`. How should we handle it?

We currently have a priority list implemented on https://github.com/opentrials/processors/blob/master/processors/base/writers/trial.py#L46-L61 to decide which source would become the "primary" trial. This should be replicated here so, in this example, we would pick the trial from `nct` instead of the one from `pfizer`. This would require some refactoring, though, possibly extracting this priority list to a single place (maybe https://github.com/opentrials/processors/blob/master/processors/base/config.py) and reusing it here and in the trial writer.

However, as this is a corner case and we don't have much more time to work on this PR, I'm happy to simply use the first trial from a different source (i.e. `SELECT * FROM trials WHERE public_title = :public_title AND source_id != :source_id LIMIT 1`). We can create an issue for this case and fix it later.

arthurSena: Ok, now it's clear. 
We already made the commit that fixes the first case you just explained.

processors/base/helpers/__init__.py
arthurSena: You're right, we checked this on the docker and it doesn't work. 
We will tackle this issue and try to find a way to work around

arthurSena: As there is no git repository in the docker, we thought about using the last commit hash found in the database if it is running in the docker container. Is that an acceptable approach ?

vitorbaptista:
Will be merged on https://github.com/opentrials/processors/pull/123

PR #118, opened 2017-02-21
[#685] Treat document_categories as static data

georgiana-b:
opentrials/opentrials#685


processors/fda_dap/processor.py
```diff
@@ -45,16 +45,15 @@ def __init__(self, conf, conn):
 
     def process_record(self, record, source_id):
         fda_approval = self._write_fda_approval(record)
-        document_category_id = self._upsert_document_category()
-
+        doc_category_id = self._conf['DOCUMENT_CATEGORIES']['other']
```
victor-nitu, line 48: Why the variable rename? (`document_category_id` -> `doc_category_id`)

victor-nitu (approved):
I'm okay with it, although still not happy about keeping IDs in the code. But that's definitely an improvement and makes it a lot easier for collecting them dynamically in the future. I'd say go for it.

processors/fda_dap/processor.py
georgiana-b: I also changed the way the id was retrieved which made the row too long.

PR #119, opened 2017-02-24
[#628] Feature/data contributions processor

georgiana-b:
opentrials/opentrials#628

This is code is compatible with the other parts of OpenTrials as they are now.

I removed the `contrib` processor because the functionality is mostly covered by the `data_contributions` processor and in the OpenTrials Explorer (uploading to S3). The only thing not currently covered is processing archives because we can't know what `document_category` each document inside them has.

There are [87 documents](https://query.opentrials.net/queries/204) generated by the `contrib` processor. I queried them [by `name`](https://github.com/opentrials/processors/blob/master/processors/contrib/processor.py#L92) because the `source_url` was [dynamically created](https://github.com/opentrials/processors/blob/master/processors/contrib/processor.py#L159) and hard to deduce.
These 87 documents seem fine and the only change I can think of is to update their `source_id` to `contrib`.


processors/data_contributions/processor.py
```diff
@@ -0,0 +1,121 @@
...
+                    id=document['document_category_id'])
+                document.update({
+                    'source_id': source_id,
+                    'name': doc_category['name'] if doc_category else None,
+                })
+                document_id = base.writers.write_document(conn, document)
+
+                # Update trial relation
+                conn['database']['trials_documents'].delete(document_id=document_id)
+                trial_rel = {
```
vitorbaptista: Unless the word is very long, avoid abbreviations, as they can reduce readability. Reading the rest of the code I understand that `trial_rel` means `trial_relation`, but it's not instantly obvious when reading the name. Also, we already have a name for this: `trials_documents` (or, in singular, `trial_document`), so it's better to keep consistent naming.

As this variable is only used on the following `.insert()` call, you could even write it directly in the method call (i.e. `.insert({ ... })`).

processors/data_contributions/processor.py
```diff
@@ -0,0 +1,121 @@
...
+        content_type = data_request.headers['Content-Type']
+
+        # Check if the document is not an archive
+        if not any(ctype in content_type for ctype in ['-tar', '-gtar', 'zip', '7z', 'zlib']):
+            conn['database'].begin()
+            try:
+
+                # Upsert document
+                doc_category = conn['database']['document_categories'].find_one(
+                    id=document['document_category_id'])
```
vitorbaptista: This is something I'd like to setup linting to catch, but while we don't have an automatic linter, please leave the last `)` in the next line, like:

```python
conn['database']['document_categories'].find_one(
    id=document['document_category_id']
)
```

processors/data_contributions/processor.py
```diff
@@ -0,0 +1,121 @@
...
+import logging
+from .. import base
+from . import extractors as extractors_module
+logger = logging.getLogger(__name__)
+
+
+def process(conf, conn):
+    def convert_ids(contrib):
+        converted_ids = {key: val.hex for key, val in contrib.items()
+                        if val.__class__.__name__ == 'UUID'}
```
vitorbaptista: In general, instead of checking for the class itself, it's better to check for the method you want. In this case, you could do `hasattr(val, 'hex')`.

tests/processors/data_contributions/test_processor.py
```diff
@@ -0,0 +1,125 @@
...
+        contrib.update(contrib_attrs)
+        conn['explorer']['data_contributions'].update(contrib, ['id'])
+        processor.process({}, conn)
+
+        updated_contrib = conn['explorer']['data_contributions'].find_one(id=valid_data_contrib)
+        assert updated_contrib['document_id'] is None
+
+
+@pytest.fixture
+def valid_data_contrib(conn, data_contribution, document_category, trial):
```
vitorbaptista: Why is it `valid_data_contrib`? Is the regular `data_contribution` fixture invalid?

processors/data_contributions/processor.py
```diff
@@ -0,0 +1,121 @@
...
+        data_contrib = convert_ids(data_contrib)
+        _remove_document(conn, data_contrib)
+
+
+def _process_document(conn, contrib, extractors, source_id):
+    document_id = None
+    document = extractors['extract_document'](contrib)
+    data_request = _request_data_contribution(document['source_url'])
+    if data_request:
+        content_type = data_request.headers['Content-Type']
```
vitorbaptista: It feels overkill to download a potentially large document just to check its content type. It would be better to either validate the URL's file extension (as in `.pdf`, `.zip`, etc) or at least use a HEAD request instead of a GET, so we don't actually download the file. I think it's fine to check the file extension only.

Readability could be improved with better naming and extracting a method. For example, you could do something like:

```python
def _process_document(conn, contrib, extractors, source_id):
    if not _document_has_valid_format(document):
        return
    # Process the document here


def _document_has_valid_format(document):
    # Check if the format is valid
```

I think it would also be better to pass the document itself to the `_process_document()` function, instead of sending in the `extractors`. This way all extracting happens on the `process()` function. The method signature would become `_process_document(conn, document, data_contribution, source_id)`

processors/data_contributions/processor.py
```diff
@@ -0,0 +1,121 @@
...
+        data_contrib = convert_ids(data_contrib)
+        trial = conn['database']['trials'].find_one(id=data_contrib['trial_id'])
+        if not trial:
+            continue
+        _process_document(conn, data_contrib, extractors, source_id)
+
+    # Remove documents from data_contributions
+    for_removal = """SELECT * FROM data_contributions
+                     WHERE approved=false
+                     AND document_id IS NOT NULL;"""
```
vitorbaptista: It's harmless and no problem in leaving it in, but the trailing semicolon is unnecessary here.

Please use single quotes and format the queries in separate lines, as:
```python
for_removal = '''
    SELECT * FROM data_contributions
    WHERE approved = false
    AND document_id IS NOT NULL
'''
```

processors/data_contributions/extractors.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+
+def extract_source():
+    return {
+        'id': 'contrib',
```
vitorbaptista: No need to abbreviate here. Use `data_contribution` or simply `contribution` to improve readability.

vitorbaptista (changes requested):
Please check the inline comments. Also, the processor is following the pattern where we try to process as much as possible, swallowing exceptions as we go. This is OK-ish for this processor, as we could argue that we don't want the processor to stop if a single a data contribution has an invalid `trial_id` (caused by human error). However, we do want to stop if there was a code error.

Because of this, I'd guess it would be better to swallow `SQLAlchemyError` instead of `Exception`, as it'll let through non-DB related exceptions.

tests/processors/data_contributions/test_processor.py
georgiana-b: Yes, a valid data contribution passes [this query](https://github.com/opentrials/processors/pull/119/commits/707b0f5207343e37bda261b5882058e0f3ba2c6e#diff-afa5df5f8a3a258e5bfa03a15f836566R29).

georgiana-b:
@vitorbaptista Do I have to add `data_contributions` processor to [docker-compose.yml](https://github.com/opentrials/processors/blob/master/docker-compose.yml)?

georgiana-b:
@nightsh Looks like `codacy` needs some configuration to tolerate `pytest` rules.

victor-nitu:
@georgiana-b will do, that's why I also added it on a py repository to see if it goes along well with our code base. We can just ignore it until we're sure it plays after our rules (it's configurable enough, thankfully).
Just some toggles and switches to select individual rules to apply (or not). Easy peasy.

victor-nitu:
@vitorbaptista can we merge this?

tests/processors/data_contributions/test_processor.py
vitorbaptista: As it looked like the same `data_contribution` fixture, only with an actual `data_category`, I removed it and added the `data_category` to the regular `data_contribution` fixture.

vitorbaptista (approved):

PR #120, opened 2017-03-02
[#656] Process unregistered trials

georgiana-b:
opentrials/opentrials#656

As described in the issue above, an unregistered `trial` and its corresponding `record` are created if the PubMed record doesn't have recognizable `registry_ids`(e.g. NCT ID). Otherwise, a `document` of category `Results` is created and attached to the `trial` referenced in `registry_ids`.
If a PubMed record is updated with `registry_ids` after an unregistered `trial` has been created from it, the `trial's`  pubmed `record` will be removed. **To remove the unregistered `trial` itself, the `trial_remover` processor has to be run after `pubmed` processor is finished.**

We currently [have 442 trials from PubMed records](https://query.opentrials.net/queries/205). These were created by the [`pubmed_linker` processor](https://github.com/opentrials/processors/compare/feature/process-unregistered-trials?expand=1#diff-98246c30aed101571e9bb8e28829f8a6L48). These `trials` don't have `records` and they [will be removed at the next run of `trials_remover` processor](https://github.com/opentrials/processors/compare/feature/process-unregistered-trials?expand=1#diff-96540bf4c950e81a6cef60605a8b2983L31).

I removed the `pubmed_linker` and `hra_linker` processors because they were doing the same thing. I moved the linkage of `publications` and `trials` in the `publication` base processor, as we do in the other base processors.


processors/pubmed/extractors.py
```diff
@@ -42,6 +52,38 @@ def extract_publications(record):
...
+
+
+def extract_document_category(record):
+    return base.config.DOCUMENT_CATEGORIES['journal_article']
+
+
+def extract_trial(record):
+    pubmed_id = base.helpers.safe_prepend('PUBMED', record['pmid'])
+    identifiers = []
+    ids = base.helpers.find_list_of_identifiers(pubmed_id)
```
victor-nitu: Use named parameter here to increase readability: `base.helpers.find_list_of_identifiers(text=pubmed_id)`

processors/pubmed/extractors.py
```diff
@@ -42,6 +52,38 @@ def extract_publications(record):
...
+    if len(ids) == 1:
+        identifiers = ids[0]
```
victor-nitu: This will cast the `identifiers` list into a string. Not sure if it's the intended behaviour.

processors/pubmed/processor.py
```diff
@@ -4,13 +4,67 @@
...
+            # Publication is not/no longer an unregistered trial but a results document
+            else:
+                delete_record = "DELETE FROM records WHERE identifiers @> '%s'"
+                delete_record = delete_record % json.dumps(unregistered_trial['identifiers'])
+                conn['database'].query(delete_record)
+                for identifiers in publication['registry_ids']:
+                    trial = base.helpers.find_trial_by_identifiers(
+                        conn, identifiers=identifiers
+                    )
+                    if trial:
```
victor-nitu: And what if we don't have any? Say publication is updated and **one** valid identifier is added. But we didn't collect it, for some reason. It's not clear what should be happening then.

victor-nitu (changes requested):
See comments. Not necessarily requested changes, more like "requested explanations" in order to push it forward.

processors/pubmed/processor.py
georgiana-b: If a publication has `registry_ids` but the `trial` is not found the results `document` will not be created. 
However, it might be that the `publication` *did not* have `registry_ids` before which means we have an unregistered trial for it. So first [we delete the `record`](https://github.com/opentrials/processors/blob/525b68c2f5ea7a1f8846e3ed05814a7cdd9ed801/processors/pubmed/processor.py#L46-L48) of the unregistered trial which will cause the unregistered trial to be removed by the `trial_remover` processor.

victor-nitu (approved):

victor-nitu:
Alright, go for it. just fix the conflicts and you can do the merge :+1:

georgiana-b:
@nightsh After taking a second look, I think it would be better to extract [the logic to process unregistered trials](https://github.com/opentrials/processors/blob/525b68c2f5ea7a1f8846e3ed05814a7cdd9ed801/processors/pubmed/processor.py#L25) in a different processor.

georgiana-b:
@nightsh I extracted the logic for processing unregistered trials in a  different processor. If you have some time, please take a look.

processors/base/processors/publication.py
```diff
@@ -5,14 +5,13 @@
 from __future__ import unicode_literals
 
 import logging
+from sqlalchemy import exc as sqlalchemy_exceptions
```
vitorbaptista: The main reason I see for using `from X import Y as Z` is to make the words smaller, but that's not the case here. I tend to avoid importing like this unless it's much more convenient, as it makes it harder IMHO to track where things come from in the code, and makes the code less "greppable". I'd rather change this to `import sqlalchemy.exc` and use `sqlalchemy.exc.DBAPIError`.

processors/base/writers/trial.py
```diff
@@ -42,7 +42,7 @@ def write_trial(conn, trial, source_id, record_id):
 
     # Decide primary
     is_primary = False
-    priority = ['nct', 'euctr', 'isrctn']
+    priority = ['nct', 'euctr', 'isrctn', 'pubmed']
```
vitorbaptista: I think PubMed shouldn't be here, as it's the least important source we have (i.e. if there's any other source for the same trial, we prefer that instead of PubMed). If it's the only trial's source, it'll become the primary trial anyway, regardless of the priority list.

processors/pubmed_unregistered_trials/extractors.py
```diff
@@ -0,0 +1,33 @@
...
+def extract_document_category(record):
+    return base.config.DOCUMENT_CATEGORIES['journal_article']
+
+
+def extract_trial(record):
+    pubmed_id = base.helpers.safe_prepend('PUBMED', record['pmid'])
+    identifier = None
+    ids = base.helpers.find_list_of_identifiers(pubmed_id)
+    if len(ids) == 1:
+        identifier = ids[0]
```
vitorbaptista: This is a bit strange. The question that comes to my mind while reading this is what happens if `ids` have length > 1. Is that possible at all?

I think you're calling `find_list_of_identifiers()` here to validate the `pubmed_id`. If that's the case, doing something like:

```python
pubmed_id = base.helpers.safe_prepend('PUBMED', record['pmid'])
identifiers = base.helpers.clean_identifiers({
    'pubmed': pubmed_id,
})
```

Is clearer IMHO.

processors/pubmed_unregistered_trials/processor.py
```diff
@@ -0,0 +1,70 @@
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import json
+import logging
+from sqlalchemy import exc as sqlalchemy_exceptions
```
vitorbaptista: Same as https://github.com/opentrials/processors/pull/120/files#r108697191

processors/pubmed_unregistered_trials/processor.py
```diff
@@ -0,0 +1,70 @@
...
+from .. import base
+from . import extractors as extractors_module
+from ..pubmed_publications import extractors as related_extractors_module
```
vitorbaptista: I think the name `related_extractors_module` is confusing. Related to what? I think `pubmed_publications_extractors` would be clearer.

On this note, I know that you're following the existing pattern of adding `_module` to the `extractors` name, but it feels redundant. For example, we don't use this pattern when importing `base`, even though it's also a module. I'd rather remove all `_module` suffixes instead of following it. We can do this in the end, before merging this PR, or you can do now, whatever you prefer.

processors/pubmed_unregistered_trials/processor.py
```diff
@@ -0,0 +1,70 @@
...
+    )
+    related_extractors = base.helpers.get_variables(
+        related_extractors_module, lambda x: x.startswith('extract_')
+    )
+    source = related_extractors['extract_source'](None)
+    source_id = base.writers.write_source(conn, source)
+
+    for record in base.helpers.iter_rows(conn, 'warehouse', 'pubmed', orderby='meta_id'):
+        publication = related_extractors['extract_publication'](record)
+        unregistered_trial = extractors['extract_trial'](record)
```
vitorbaptista: We don't know if it's an unregistered trial at this point, as that depends on `publication['registry_ids']` existing or not, so it's better to simply name it `trial`

processors/pubmed_unregistered_trials/processor.py
```diff
@@ -0,0 +1,70 @@
...
+                )
+                base.writers.write_record(
+                    conn, record, source_id, trial_id, unregistered_trial, is_primary
+                )
+
+            # Publication is not/no longer an unregistered trial but a results document
+            else:
+                delete_record = "DELETE FROM records WHERE identifiers @> '%s'"
+                delete_record = delete_record % json.dumps(unregistered_trial['identifiers'])
+                conn['database'].query(delete_record)
```
vitorbaptista: Instead of generating the SQL as is, do:

```python
query = 'DELETE FROM records WHERE identifiers @> :identifiers'
conn['database'].query(query, identifiers=unregistered_trial['identifiers'])
```

This is a better pattern because SQLAlchemy takes care of safely converting the parameters for us.

processors/pubmed_unregistered_trials/processor.py
```diff
@@ -0,0 +1,70 @@
...
+                    trial = base.helpers.find_trial_by_identifiers(
+                        conn, identifiers=identifiers
+                    )
+                    if trial:
+                        doc_category_id = extractors['extract_document_category'](record)
+                        document = extractors['extract_document'](record)
+                        document.update({
+                            'source_id': source_id,
+                            'document_category_id': doc_category_id,
+                        })
```
vitorbaptista, line 57: It feels like the `extract_document` extractor should define the `source_id` and `document_category_id`.

vitorbaptista: BTW, I understand we're only ever returning a single document, but it feels somewhat odd that everywhere else we use `extract_documents` except in this processor.

processors/pubmed_unregistered_trials/processor.py
```diff
@@ -0,0 +1,70 @@
...
+                        document_id = base.writers.write_document(conn, document)
+                        base.writers.write_trial_relationship(
+                            conn, 'document', document, document_id, trial['id']
+                        )
+        except sqlalchemy_exceptions.DBAPIError:
+            conn['database'].rollback()
+            base.config.SENTRY.captureException(extra={
+                'record': record,
+            })
```
vitorbaptista: I agree with swallowing exceptions on the `data_contributions` processors, but why do the same here? Swallowing exceptions have the bad side-effect that our processors will pass on Airflow, so it won't notify us, retry or stop the next steps.

tests/processors/base/processors/test_publications.py
```diff
@@ -0,0 +1,47 @@
...
+        record_attrs = conn['database']['records'].find_one(id=record)
+
+        identifier = {'nct': 'NCT00020500'}
+        pubmed_attrs['article_title'] = ('This publication is related to %s' % identifier)
+        record_attrs.update({
+            'trial_id': trial,
+            'identifiers': identifier,
+        })
+        conn['warehouse']['pubmed'].update(pubmed_attrs, ['pmid'])
+        conn['database']['records'].update(record_attrs, ['id'])
```
vitorbaptista: An easier way of doing this pattern of getting a fixture and changing something in it is:

```python
def test_foo(self, pubmed_record):
    conn['warehouse']['pubmed'].update(
        {
            'pmid': pubmed_record,
            'article_title': 'This publication is related to %s' % identifier,
        },
        ['pmid']
    )
```

In other words, you don't need to query the DB to get the data, unless you need it for something else. This means less code and a quicker test :+1:

tests/processors/base/processors/test_publications.py
```diff
@@ -0,0 +1,47 @@
...
+    def test_creates_publication(self, conn, extractors, pubmed_record):
+        pubmed_attrs = conn['warehouse']['pubmed'].find_one(pmid=pubmed_record)
+        process_publications(conn, 'pubmed', extractors)
+
+        publication = conn['database']['publications'].find_one(
+            source_url=pubmed_attrs['meta_source']
+        )
+        assert publication is not None
+
+
```
vitorbaptista, line 23: You're using 2 newlines across the code, which I guess came from lint complaining about it. However, lint only requires 2 newlines for functions outside of a class. If the functions are inside the class, it requires a single newline. This is passing because we're not linting the tests code.

I'm happy to leave this as is, as it'll eventually be fixed when we do https://github.com/opentrials/opentrials/issues/701. Just letting you know so new code don't follow this pattern.

tests/processors/pubmed_publications/test_extractors.py
```diff
@@ -0,0 +1,37 @@
...
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+
+import collections
+import processors.pubmed_publications.extractors as extractors
+
+
+class TestPubmedPublicationExtractor(object):
```
vitorbaptista: Publication -> Publications

processors/base/helpers/__init__.py
```diff
@@ -90,6 +90,7 @@ def clean_identifiers(identifiers):
         'ntr': r'^NTR',
         'pactr': r'^PACTR',
         'per': r'^PER',
+        'pubmed': r'^PUBMED\d',
```
vitorbaptista: I think I would either use `r'^PUBMED\d+'` or keep it as `r'^PUBMED'`

vitorbaptista (changes requested):
This is an impressive work, congratulations! :clap:

I added a few comments, but nothing too important. I think the implementation is solid, and am really happy to see this merged. Please take a look at the comments so we can move forward.

processors/pubmed_unregistered_trials/processor.py
georgiana-b: `source_id` [comes from within the processor](https://github.com/opentrials/processors/blob/028150961ce1b5686ec9040201f9bb5e710a8774/processors/pubmed_unregistered_trials/processor.py#L24). I defined `document_category_id` here just to keep [the convention](https://github.com/opentrials/processors/blob/028150961ce1b5686ec9040201f9bb5e710a8774/processors/base/processors/trial.py#L64).

georgiana-b: As for `extract_document` I would like to keep the convention but I don't agree with it. Even though the function is called `extract_documents` it extracts only 1 document in all processors that have it. 
I think we should rename the function `extract_document` in all other processors and avoid doing [this](https://github.com/opentrials/processors/blob/028150961ce1b5686ec9040201f9bb5e710a8774/processors/gsk/extractors.py#L128). What do you think?

processors/pubmed_unregistered_trials/processor.py
```diff
@@ -6,43 +6,44 @@
...
 
             # Publication is not/no longer an unregistered trial but a results document
             else:
-                delete_record = "DELETE FROM records WHERE identifiers @> '%s'"
-                delete_record = delete_record % json.dumps(unregistered_trial['identifiers'])
-                conn['database'].query(delete_record)
+                delete_record_query = 'DELETE FROM records WHERE identifiers @> :identifiers'
+                conn['database'].query(
+                    delete_record_query,
+                    identifiers=json.dumps(potential_trial['identifiers'])
```
georgiana-b: @vitorbaptista I had to use `json.dumps` because passing the dict directly raises
 ` ProgrammingError: (psycopg2.ProgrammingError) can't adapt type 'dict' [SQL: 'DELETE FROM records WHERE identifiers @> %(identifiers)s'] [parameters: {'identifiers': {u'pubmed': u'PUBMED23571166'}}]
`

processors/pubmed_unregistered_trials/processor.py
vitorbaptista: Regarding `extract_document` vs `extract_documents`, I agree to rename them all to `extract_document`. Keep this PR as is and please create a new issue for this renaming.

vitorbaptista (approved):

vitorbaptista:
Great job! :beers:

PR #122, opened 2017-03-21
[#628] Ignore record-level errors for data contributions

georgiana-b:
opentrials/opentrials#628


vitorbaptista (changes requested):
Instead of removing the `.extra_context()` calls and adding `captureException()` in their place, it's better to keep the `.extra_context()` and simply call `raise` inside the `sqlalchemy_exceptions.DBAPIError` catch. This way we keep the context for Sentry not only when `sqlalchemy_exceptions.DBAPIError` is thrown, but on any exception.

georgiana-b:
@vitorbaptista I reverted the changes to keep the `.extra_context()` part. However, I had to keep `captureException()` instead of  `raise` because `raise` would stop the processor.

vitorbaptista (approved):

PR #123, opened 2017-04-03
[#661] Deduplicate trials via title

vitorbaptista:
This PR continues the work from https://github.com/opentrials/processors/pull/115 (I couldn't push to that branch, so needed to start a new one). Please check that for more information.

Before this is merged, we need to uncomment the `deploy` lines on `.travis.yml`

Connects opentrials/opentrials#661


vitorbaptista:
Please, whoever review this check https://github.com/opentrials/api/pull/153 also

vitorbaptista:
Couldn't find the codacy report, but that shouldn't block merging this IMHO

processors/base/helpers/__init__.py
```diff
@@ -250,27 +250,94 @@ def find_trial_by_identifiers(conn, identifiers, ignore_record_id=None):
...
+        if trial_ids:
+            return conn['database']['trials'].find_one(id=trial_ids)
+
+
+def find_trial_by_public_title(conn, public_title, ignore_source_id, ignore_record_id=None):
+    QUERY = (
+        'SELECT trial_id FROM records'
+        ' WHERE public_title = :public_title'
+        ' AND source_id != :source_id'
+        ' AND id != :record_id'
```
victor-nitu, line 286: As discussed, regarding `ignore_record_id` interpolation vs. conditional query building: it doesn't make any practical difference. Not sure about the readability being improved in any case, this is shorter, the other is more explicit. I suppose the only thing that matters is to be consistent in the two functions.

I'd add a short line comment to make things clearer tho, something like `# This avoids breaking the query when ignore_record_id is not supplied`

processors/base/helpers/__init__.py
```diff
@@ -250,27 +250,94 @@ def find_trial_by_identifiers(conn, identifiers, ignore_record_id=None):
...
+            source_id,
+            ignore_record_id
+        )
+
+        if not trial_found:
+            trial_found = find_trial_by_public_title(conn, trial['public_title'], source_id)
+
+        if trial_found:
+            deduplication_method = 'public_title'
+
```
victor-nitu: I would suggest:

* find by identifiers
  * if not found, find by public title
* if none of the above have results, try them again by using the `ignore_record_id`
* whenever anything is found, return (so we'd get rid of the misleading multiple indentation for code that has the same purpose)
  * alternatively, you could set a `trial_found` "flag variable" and run everything inside a `while` loop, or something similar – at the end of it, you just return it

However, this is not a crucial matter. It would probably increase the readability of code a bit.

processors/base/writers/record.py
```diff
@@ -70,25 +69,32 @@ def write_record(conn, record, source_id, trial_id, trial, is_primary):
...
-    return obj['id']
+    if helpers.validate_remote_url(record['source_url']):
+        try:
+            conn['database'].begin()
+
+            if record['is_primary']:
+                conn['database']['records'].update(
+                    {
+                        'trial_id': record['trial_id'],
+                        'is_primary': False,
```
victor-nitu: is this a 3 spaces indent?

processors/base/writers/trial.py
```diff
@@ -28,77 +32,106 @@ def write_trial(conn, trial, source_id, record_id):
...
+        'created' if create else 'updated', trial_attrs['identifiers'])
+
+    return trial['id'], is_primary
+
+
+def _write_deduplication_log(conn, trial_id, record_id, method):
+    latest_log = conn['database']['trial_deduplication_logs'].find_one(
+        trial_id=trial_id,
+        record_id=record_id,
+        order_by='-created_at'
```
victor-nitu: I'll say it again: nice one :nerd_face:

victor-nitu (commented):
Hey, I wrote some comments on the diff. They're not blocking the merge, as the functionality isn't affected, but they might tackle readability a bit. Feel free to follow or discard them.

I haven't run it, so I'd say let's get it in today, together with the other PRs in the family, and start a big processing task over the weekend ;)

processors/base/writers/record.py
vitorbaptista: Looks fine to me :thinking:

PR #124, opened 2017-04-05
[#643] Overwrite trials and records with the new data

vitorbaptista:
_NOTE:_ This depends on https://github.com/opentrials/processors/pull/123 being merged first.

For example, consider the case that some data is removed from the source (e.g.
it used to have "target_sample_size" but doesn't anymore). If we don't clear
the row when updating the data, this field wouldn't be cleared, and so we'd end
up with a mix of old and new data.

Fixes opentrials/opentrials#643


vitorbaptista:
This PR isn't needed anymore, as I wrote a better test for the case we're checking on #661 (e.g. https://github.com/opentrials/processors/blob/e26fb217d1e7ab2cd3366ccb2852fecc0168bbb3/tests/processors/base/writers/test_trial_and_record.py#L244-L266)

PR #126, opened 2017-04-11
[#173] Extract age range

vitorbaptista:
opentrials/opentrials#173


processors/base/helpers/__init__.py
```diff
@@ -413,3 +413,65 @@ def get_canonical_location_name(location):
...
+        'hour',
+        'minute',
+    ]
+    if age_unit.rstrip('s') in AGE_UNITS:
+        unit = age_unit.rstrip('s')
+        if age_num != 1:
+            unit += 's'
+        return '{} {}'.format(age_num, unit).title()
+    else:
+        return 'Unknown'
```
georgiana-b: Why does wrong unit mean "Unknown" whereas wrong number means ValueError?
IMO this is no different from the above situation where `age_num` doesn't fit the pattern so it would seem fair to treat them the same.

processors/pfizer/extractors.py
```diff
@@ -93,3 +97,18 @@ def extract_organisations(record):
...
     return persons
+
+
+def _extract_age_range(record):
+    age = (record.get('age_range') or '').strip().lower()
+    min_age, max_age = None, None
+    if age.endswith(' and older'):
+        min_age = age[:-len(' and older')]
+        max_age = 'N/A'
+    elif len(age.split(' - ')) == 2:
```
georgiana-b: Wouldn't it be better to normalize the spaces around `-` out before splitting just to cover for cases like `12-13`?

tests/processors/pfizer/test_extractors.py
```diff
@@ -0,0 +1,65 @@
...
+        })
+
+        trial = extractors.extract_trial(stub_record)
+
+        assert trial['age_range'] == {
+            'min_age': expected_min_age,
+            'max_age': expected_max_age,
+        }
+
+    def test_minimum_and_maximum_age_are_unknown_by_default(self, stub_record):
```
georgiana-b: Isn't this already covered [in the test above](https://github.com/opentrials/processors/blob/52ac6749d92c7dfe57ca6ded32c7af05b3c4d238/tests/processors/pfizer/test_extractors.py#L30)?

georgiana-b (changes requested):

PR #127, opened 2017-04-11
Add new `journal` type

victor-nitu:
This will only apply to PubMed at the moment.

Related to opentrials/opentrials#708, in which we need to find items coming from a register or a journal publication.


vitorbaptista:
This was already merged on https://github.com/opentrials/processors/pull/125

PR #128, opened 2017-04-12
[#711]  Betamax intergration

georgiana-b:
Everything appears to work ok when there is internet connection: cassettes are created for each test; tests pass.
However, when tests run offline, they all fail because `requests` library raises a `ConnectionError`. From what I understand, if the betamax integration would have been successful, that error shouldn't be raised. I tried to figure out why tests fail offline but I failed so any help would be appreciated.


tests/conftest.py
```diff
@@ -29,6 +31,20 @@
...
+# Betamax config
+
+with betamax.Betamax.configure() as cfg:
+    cfg.cassette_library_dir = 'tests/cassettes/'
+
+    record_mode = 'none' if os.environ.get('CI') else 'once'
+    cfg.default_cassette_options['record_mode'] = record_mode
+    cfg.default_cassette_options['match_requests_on'] = [
+        'uri',
+        'method',
```
georgiana-b, line 44: I just realized this is probably why the tests fail: the match on `method` fails because the [processor makes a `HEAD` request] while betamax made a `GET` request.

georgiana-b: Update: The problem has not been solved by removing the match on `method`.

vitorbaptista:
@georgiana-b Please check my fix on https://github.com/opentrials/processors/pull/128/commits/9ef1a333c1d08873d7ad006ef8078ffe65a21599. The issue was that `betamax` doesn't mock the `requests` automatically, so you have to do some sort of dependency injection to make the code use `betamax` instead of `requests`.

My first try was simply exposing the `requests` method as a method param, so we could pass `betamax` when testing. For example, on https://github.com/opentrials/processors/blob/e26fb217d1e7ab2cd3366ccb2852fecc0168bbb3/processors/cochrane_reviews/processor.py#L150-L167 we have:

```python
def scrape_pubmed_id(reference):
    """Try to find PUBMED ID for reference on PubMed"""
    base_url = 'https://eutils.ncbi.nlm.nih.gov/entrez/eutils/esearch.fcgi'
    # ...
    search_result = requests.get(base_url, params=search_params)
    # ...
```

So I changed that to:

```python
def scrape_pubmed_id(reference, get_url=requests.get):
    """Try to find PUBMED ID for reference on PubMed"""
    base_url = 'https://eutils.ncbi.nlm.nih.gov/entrez/eutils/esearch.fcgi'
    # ...
    search_result = get_url(base_url, params=search_params)
    # ...
```

And in the tests I would simply call it as `scrape_pubmed_id(reference, betamax_session.get)`. This works fine, but requires changing the code which I didn't like much. As Python is a very dynamic language, I thought we could do the same using the `mock` library. Unfortunately, I can't change the entire `requests` module for `betamax_session`, so I had to patch each method with `mock.patch('requests.get', side_effect=betamax_session.get)`. This gets repetitive quickly, so the last step was to move this mocking to a single place.

I remembered pytest's yield fixtures and thought that might work. The idea was to mock the `requests` methods and then yield to the test function. There's no need to pass anything to it, simply yield the execution thread, so it would run inside the context with the mocked `requests`. The current code is equivalent to:

```python
def enable_betamax(betamax_session):
    with mock.patch('requests.get', side_effect=betamax_session.get):
        test_method_were_running()
```

With this fixture, enabling betamax is simply adding the `@pytest.mark.usefixtures('enable_betamax')` decorator. This is not how I've done things in the collectors, but I like it. Please review the code and, if everything's alright, let's squash the commits and merge it :beers:

vitorbaptista (approved):

georgiana-b:
@vitorbaptista Very neat solution :+1: Let's merge it!

PR #131, opened 2017-04-23
[#757] Default is_primary to false instead of previous value in record

georgiana-b:
opentrials/opentrials#757

To see what was causing the bug see my comment :point_down:


processors/base/writers/trial_and_record.py
```diff
@@ -126,7 +126,7 @@ def _write_record(conn, trial, record_id, source_id=None, source_url=None, is_pr
         'trial_id': trial['id'],
         'source_id': source_id or record['source_id'],
         'source_url': source_url or record['source_url'],
-        'is_primary': is_primary or record.get('is_primary'),
```
georgiana-b: New records [don't have `is_primary` property](https://github.com/opentrials/processors/blob/5c21995f82a0d297618f75e7c598bf83ebfcf446/processors/base/writers/trial_and_record.py#L118) at this point in the code so for them `record.get('is_primary')` returns `None`.

The bug is caused by the fact that `is_primary or None` will return `None` if `is_primary` is `False`. Here is some explanatory console output:

```python

In [1]: val = False or None
In [2]: type(val)
Out[2]: NoneType

In [3]: val = None or False
In [5]: val
Out[5]: False

In [8]: val = False or True
In [9]: val
Out[9]: True
```

As you can deduce from this last example, this will also cause records that are no longer primary (`is_primary=False`) to take on their previous value (`True`).
The value of `is_primary` is something [we calculate](https://github.com/opentrials/processors/blob/5c21995f82a0d297618f75e7c598bf83ebfcf446/processors/base/writers/trial_and_record.py#L45) and it should always have the latest value so I just deleted `record.get('is_primary')`.

vitorbaptista: Nice catch! I think there're 2 more changes we should do:

1. Add a test that triggers this bug, making sure that it's solved (and, more importantly, that it doesn't appear again in the future)
2. The `_write_record()` parameters `source_id`, `source_url` and`is_primary` shouldn't be optional anymore. This method is only used on this file, and we always set those parameters, so it's both useless to have them optional, and confusing.

I think part of this confusion is because we used to support partial updates, where if a field wasn't set in the new data, we would keep what we had in the DB. However, after https://github.com/opentrials/opentrials/issues/643, this behaviour changed. We don't do partial updates anymore. If a field isn't set in the new data, we set it as `null`. We still have a few leftover code of the previous behaviour, though. For example, `'source_id': source_id or record['source_id']` basically means "If the user haven't set a `source_id`, use the current one in the record". This is a partial update. The correct behaviour should be to require the user to define a `source_id` (as we can't set it as `None` because it's required).

So, on my second point up there, the changes would be to require all parameters of `_write_record()` to be set and remove change the pattern `source_id or record['source_id']` to simply `source_id` (same for `source_url`).

Makes sense?

vitorbaptista (changes requested):

tests/processors/base/writers/test_trial_and_record.py
```diff
@@ -284,16 +284,23 @@ def test_writes_record_with_valid_url(self, conn, trial, nct_source):
...
+            trial_and_record_writer._write_record(conn_mock, trial_attrs, record, None, None, False)
+
+            conn_mock['database']['records'].upsert.assert_called()
 
-        trial_and_record_writer._write_record(conn, trial_attrs, record, None, None)
+            upsert_call_args = conn_mock['database']['records'].upsert.call_args
+            upsertted_record = upsert_call_args[0][0]
+            assert upsertted_record['is_primary'] is False
+            assert upsertted_record['source_url'] is None
+            assert upsertted_record['source_id'] is None
```
vitorbaptista: These assertions never run, because `_write_record()` does raise an Exception. See:

```python
# This test pass
def test_foo():
    with pytest.raises(Exception):
        raise Exception()
        assert False  # This never gets executed

    assert True  # This does
```

tests/processors/base/writers/test_trial_and_record.py
```diff
@@ -284,16 +284,23 @@ def test_writes_record_with_valid_url(self, conn, trial, nct_source):
 
         assert conn['database']['records'].count(id=record_id) == 1
 
-    def test_doesnt_update_None_parameters(self, conn, trial, record):
+    def test_doesnt_alter_record_parameters(self, conn, trial, record):
```
vitorbaptista: I don't understand what we're testing here. It feels like the objective is to guarantee that we don't set a record's `is_primary` to `False`, and `source_url` and `source_id` to `None`. However, the actual test is that `source_url` is a valid URL. Because if you called `_write_record(conn_mock, trial_attrs, record, None, 'invalid_source_url', False)`, the exception would be different (some `sqlalchemy` exception I imagine).

The deeper problem here IMHO is trying to test the behaviour of the code, not the results of running it. In other words, this is a white box test, where a black box test is usually better. This way, the tests don't break unnecessarily if we refactor the code.

My suggestion would be to simply delete this test. I think the only param type validation that we aren't testing is checking that `source_id` isn't `None` (we already check if `source_url` is valid and `is_primary` can be `None`). However, what guarantees `source_id` can't be `None` is the database. As it doesn't change often, I don't think we need this test.

vitorbaptista (changes requested):

PR #132, opened 2017-04-24
[#702] Headers only GET request for data contributions

georgiana-b:
opentrials/opentrials#702

As explained in #702, these changes allow us to make a GET request without downloading the contents of data contributions. The body of the requests is downloaded only if `response.body` is called which betamax [does](https://github.com/sigmavirus24/betamax/blob/1f98e893473ef39390a3d72e87486e8d7ec124e2/betamax/util.py#L37) so we end up having the requests body in the cassettes even though the code doesn't download it.


georgiana-b:
@vitorbaptista Do you think we should keep this solution to avoid downloading requests body even though we can't replicate it in tests?

vitorbaptista:
@georgiana-b Yeah, that's fine :v: Thanks!

PR #135, opened 2017-08-10
Fixed typo ('lead_spondor'), added `enrollment` field handling

i1t:
* 'lead_spondor' should be written as 'lead_sponsor'
* `enrollment_anticipated` and `enrollment_actual` seem to be replaced with `enrollment` in ClinicalTrials.gov output


i1t:
Closing because of `enrollment` part (incorrect).

